PR #26, opened 2015-01-21
Compress responses

andrewgarner:
Further reading: http://robots.thoughtbot.com/content-compression-with-rack-deflater


gjvis:
:shipit:

PR #27, opened 2015-01-29
Guides served from root

andrewgarner:
- 'Articles' are now 'Guides'
- URLs are no longer prefixed and are shorter - e.g. `/articles/tax-and-pensions` is now `/tax`


features/guides.feature
```diff
@@ -0,0 +1,22 @@
...
+  As a visitor to the website
+  I want informative content on the website
+  So that I can educate myself about pensions and retirement
+
+  Scenario Outline: Load a guide
+    When I visit the <id> guide
+    Then the guide is displayed
+
+  Examples:
+    | id                        |
```
gjvis: I think "Slug" would be a more descriptive header

gjvis:
Looks good, one minor clarity thing in the feature, then :ship: :ship: :ship:

PR #32, opened 2015-01-30
Add staging environment

andrewgarner:
Introduces a new environment - staging - that inherits from production. This is primarily so that we can run a staging application with `RAILS_ENV=staging` and differentiate log streams, performance monitoring etc.

Removes any `Rails.env.production?` anti-patterns.


gjvis:
:+1: :rocket:

PR #57, opened 2015-02-09
Ignore .bundle

stevenwilkin:
For those of us using binstubs


gjvis:
Machine specific stuff belongs in the global `.gitignore`

stevenwilkin:
Good point @gjvis, I'll add it to my own config instead

PR #59, opened 2015-02-09
Refactor styleguide routes

stevenwilkin:
Remove duplicated controller and path details.

Reduces width buts adds height to the code so it's up for discussion wether or not it's a worthwhile change.


gjvis:
worth it :+1:

PR #64, opened 2015-02-10
Add extra space between front matter and markdown content.

jiannis:
Guides that do not have this extra space confuse the govspeak parser causing it to not recognise headers properly. [Setext VS Atx-style headers](http://daringfireball.net/projects/markdown/syntax#header) ?


gjvis:
:+1:

PR #81, opened 2015-02-17
Update divorce.md

raykhan66:
I changed the text below... 

"This is known as [pension attachment](http://www.pensionsadvisoryservice.org.uk/about-pensions/when-things-change/when-relationships-end/pension-earmarking)."

... to the following:

"This is known as pension attachment or sometimes [pension earmarking](http://www.pensionsadvisoryservice.org.uk/about-pensions/when-things-change/when-relationships-end/pension-earmarking)."


gjvis:
on it :shipit:

PR #83, opened 2015-02-19
Provide an empty session if CSRF is invalid

andrewgarner:
In production we have seen a few CORS requests using the HTTP `OPTIONS` verb. These are generating a 500 error because a `ActionController::InvalidAuthenticityToken` was being throw when we try and gracefully render a 404 page.

This change provides an empty session if CSRF token is invalid for the request but doesn't reset it completely. Used as default if `:with` option is not specified.


stevenwilkin:
:shipit:

PR #85, opened 2015-02-23
Run the staging smoke tests after deployment

stevenwilkin:


gjvis:
Cheap and cheerful! And also good enough for now. :ship: :ship: :ship:

PR #93, opened 2015-03-05
Update pension_types.md

raykhan66:
Added the following to the end of the sentence that states the State Pension amount:
", rising to £115.95 in April 2015"


andrewgarner:
cc/ @alexkouz @hvonhaaren

PR #129, opened 2015-03-31
Contact information

andrewgarner:
Adds contact page which includes information on complaints.

A redirect will be required once this is released.


andrewgarner:
Also worth mentioning this has been demoed to @mcornford and accepted.

gjvis:
:+1:

PR #132, opened 2015-04-01
Update tax.md

hvonhaaren:
Budget change


andrewgarner:
@hvonhaaren - do you know if/when https://www.gov.uk/tax-on-your-private-pension will be changed? It still says £1.25 million is the lifetime allowance which contradicts the text in our link.

hvonhaaren:
gov.uk will change 6 April when the changes actually comes in. However, on the top of our page it says that the info will apply from 6 April, so we should have the new lifetime allowance.

PR #136, opened 2015-04-02
Update appointments.md

alexkouz:
International number added

### DIFF
--- a/content/appointments.md
+++ b/content/appointments.md
@@ -44,6 +44,6 @@ Think about your financial circumstances in general and plans for retirement, eg
 
 ##How to book
 
-Call 0300 330 1001 between 8am and 10pm, Monday to Sunday. Calls cost the same as a normal call - if your calls are free, it's included.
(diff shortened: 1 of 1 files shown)


content/appointments.md
```diff
@@ -44,6 +44,6 @@ Think about your financial circumstances in general and plans for retirement, eg
 
 ##How to book
 
-Call 0300 330 1001 between 8am and 10pm, Monday to Sunday. Calls cost the same as a normal call - if your calls are free, it's included.
+Call 0300 330 1001 between 8am and 10pm, Monday to Sunday. Calls cost the same as a normal call - if your calls are free, it's included. Call 0203 733 3495 if you live abroad.
```
gjvis: Should'nt this be +44 203 733 3495?

andrewgarner: :+1:



Regards,


Andrew Garner

Guidance Guarantee Programme
HM Treasury

+44 7971 232 140

andrewgarner: Should we say "you're calling from abroad" not "you live abroad" - then we're inclusive to people on business or holiday etc?

We could qualify "abroad" as "outside the UK" since the main number is `+44` which is a UK number.

alexkouz: Agreed

PR #161, opened 2015-04-21
New options page

aduggin:
Waiting for feedback from Policy.
Responsive table requires thorough browser testing


andrewgarner:
Are these colours correct? Shouldn't we have one colour for "Leave your pension pot untouched"?

screen shot 2015-04-21 at 08 46 28

aduggin:
Yup - Dark blue for leave untouched (matches the bullet in the sub nav) - this is what Alex specified

andrewgarner:
When we introduce it as "Also consider" it has grey bar.

aduggin:
Good point - can you get Alex GC to have a look at the page and let me know if anything needs changing - happy to do this afternoon.

content/pension_pot_options.md
```diff
@@ -202,16 +295,15 @@ You can mix your options at different times in your retirement. For example, you
 
 Pension providers are developing new retirement income products all the time, some of which might mix these options.
 
-$E
```
andrewgarner: You can avoid having to fall back to HTML if you use Kramdown's inline attribute lists:

```markdown
{: .example--mix }
$E
**Example**

Your pension pot is £60,000. You decide to take 25% as tax-free cash which is £15,000.

{: .example__annuity}
You use £20,000 to buy an annuity. This gives you a taxable income for the rest of your life of about £900 a year.

{: .example__flexible}
You also put £25,000 into a drawdown fund and take out about £5,000 a year for about 5 years.
$E
```

The `{: .foo}` specifies attributes to add to the preceding element. So in this case we're adding a class of `example--mix` to the example block. Later we're adding classes of `example__annuity` and `example__flexible` to the paragraphs.

content/pension_pot_options.md
```diff
@@ -194,7 +284,10 @@ Here are some next steps if you’re interested in cashing in your whole pot:
 - check with your provider if you can take 25% tax free
 - make sure you know [how much tax you’ll pay](/tax) on the remaining 75%
 - if you want to reinvest the money, talk to a [registered financial adviser](http://www.fca.org.uk/register) first
+</div>
 
+<div class="option option--mix" id="mix">
+<hr class="hr-option hr-option--mix">
```
andrewgarner: You could also do this as:

```markdown
{: .hr-option .hr-option--mix }
* * *
```

content/pension_pot_options.md
aduggin: I'll update the example to this - good spot.

andrewgarner: Fixed with 0c06523

content/pension_pot_options.md
andrewgarner: Fixed with 0c06523

andrewgarner:
@guidance-guarantee-programme/developers - I've rebased. Everyone happy?

app/assets/stylesheets/components/_options-mix.scss
```diff
@@ -0,0 +1,18 @@
+.options-mix {
+  @include core-16;
+  padding-bottom: 10px;
+}
+
+@media (min-width: 420px) and (max-width: 639px), (min-width: 720px)  {
```
gjvis: magic numbers?

gjvis:
:+1: I will release

PR #171, opened 2015-05-21
Cache journey nav partial for each guide

stevenwilkin:
Rendering a guide page is the slowest part of the app:

image

Turns out half the time taken to display a guide page is spent rendering the journey navigation partial:

image

Caching this single partial dramatically speeds up rendering time.

Something to consider before this goes in is that the default Rails cache [uses the filesystem](http://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-filestore) and never expires(!)

### DIFF
--- a/app/views/guides/show.html.erb
+++ b/app/views/guides/show.html.erb
@@ -12,7 +12,9 @@
 
       <hr class="hr-thin">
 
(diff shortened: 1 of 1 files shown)


gjvis:
:+1:

PR #181, opened 2015-06-02
Only mount Jasmine in development

andrewgarner:
We have an unconventional staging environment within the application so the previous conditional was not expansive enough.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,7 +21,7 @@ Rails.application.routes.draw do
     end
   end
 
-  unless Rails.env.production?
+  if Rails.env.development?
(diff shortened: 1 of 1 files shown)


gjvis:
whoops!

nice catch.

:+1: :+1:

aduggin:
:+1:

andrewgarner:
Is there a reason why we mount the fixtures at `/spec/` but the unit tests at `/specs`?

stevenwilkin:
This is where [jasmine-jquery expects them to be](https://github.com/velesin/jasmine-jquery#installation) located

Issue #188, opened 2015-06-08
Clear cache / disable cache for test runs


gjvis:
Old versions of articles can sit in the cache, invalidating test runs.

PR #189, opened 2015-06-08
Disable caching in test environment

stevenwilkin:
A content change had caused a build to break in CI

Addresses #188


gjvis:
:+1:

PR #191, opened 2015-06-09
CAB locator features

gjvis:
Add features for locating face-to-face appointment locations, which won't be run currently as they are tagged with `@todo`.


gjvis:
ping @aduggin – we wrote these together this morning with @mcornford, what do you think?

PR #192, opened 2015-06-09
Change the meta description of the appointments page.

jiannis:
The idea is to get this description into the sitelink that Google generates. Google sitelinks are a black box but it seems that for the _“Book a free appointment”_ sitelink Google uses the meta-description of the page. All other sitelinks use the first H1 and paragraph of the page :santa: 

[“Google is the homepage”](https://gds.blog.gov.uk/2012/01/27/what-does-google-is-the-homepage-mean-irl)

book google sitelink


stevenwilkin:
:+1:

gjvis:
Makes sense to me. @mcornford you happy with this?

PR #193, opened 2015-06-09
[WIP] Search using valid postcode

stevenwilkin:
First pass at making the _Search using a valid postcode_ scenario pass.

Any opinions on having `Locations` under `lib/` ? Our intention is to keep this aspect of functionality separate from the main Rails app.

As discussed with @gjvis, we want to take an approach of initially faking things at the level of the features, and as we move forward to push the faking "further down the stack" until eventually we have the final implementation paired up with appropriate fixtures.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -1,0 +1,4 @@
+class LocationsController < ApplicationController
+  def index
+  end
+end
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -1,4 +1,5 @@
 class LocationsController < ApplicationController
   def index
+    @locations = Locations.nearest_to_postcode(params[:postcode], limit: 5)
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -9,6 +9,8 @@ Rails.application.routes.draw do
           id: guide.slug
     end
 
+    resources :locations, only: :index
+
     scope path: 'styleguide', controller: 'styleguide' do
       scope path: 'pages' do
         get 'homepage', action: 'pages_homepage'
--- a/features/step_definitions/appointment_locator_steps.rb
+++ b/features/step_definitions/appointment_locator_steps.rb
@@ -1,0 +1,7 @@
+When(/^I search for appointment locations near to a valid postcode$/) do
+  Pages::Locations.new.load(postcode: 'SW1A 2HQ')
+end
+
+Then(/^I should see the (\d+) appointment locations nearest to that postcode$/) do |number_of_locations|
+  pending # express the regexp above with the code you wish you had
+end
--- a/features/step_definitions/appointment_locator_steps.rb
+++ b/features/step_definitions/appointment_locator_steps.rb
@@ -1,7 +1,14 @@
 When(/^I search for appointment locations near to a valid postcode$/) do
-  Pages::Locations.new.load(postcode: 'SW1A 2HQ')
+  postcode = 'SW1A 2HQ'
+
+  allow(Locations).to receive(:nearest_to_postcode).with(postcode, limit: 5) do
+    1.upto(5).map { |x| Locations::Location.new("cab-#{x}") }
+  end
+
+  Pages::Locations.new.load(postcode: postcode)
 end
 
 Then(/^I should see the (\d+) appointment locations nearest to that postcode$/) do |number_of_locations|
-  pending # express the regexp above with the code you wish you had
+  expected_locations = number_of_locations.to_i.times.map { |i| "cab-#{i + 1}" }
+  expect(Pages::Locations.new.locations.map(&:name).map(&:text)).to eq(expected_locations)
 end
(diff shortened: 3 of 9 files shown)


gjvis:
hell yeah :+1:

config/routes.rb
```diff
@@ -9,6 +9,8 @@
           id: guide.slug
     end
 
+    get '/locations', to: 'locations#index'
```
gjvis: Perhaps?

```ruby
resources :locations, only: :index
```

app/controllers/locations_controller.rb
```diff
@@ -0,0 +1,5 @@
+class LocationsController < ApplicationController
+  def index
+    @locations = Locations.near_postcode(params['postcode'])
```
gjvis: can we create a card in todo for "what should we show if there is no postcode specified?"

gjvis: We're being very trusting of the `Locations` class, either here or app/views/locations/index.html.erb#2 needs an `Array` coercion
```ruby
@locations = Array(Locations.near_postcode(params['postcode']))
```

config/routes.rb
stevenwilkin: Seems a bit neater, I'd originally opted for a `scope` block but it was too much for the task at hand

app/controllers/locations_controller.rb
stevenwilkin: We **do** need to cover the _no postcode_ scenario

gjvis: haha, indeed, but as this piece of work? I reckon its a different scenario...

gjvis: oh you mean "no postcode" is the same as "invalid postcode", yeah alright makes sense!

stevenwilkin: Yep, it's a different scenario, I should have been clearer

stevenwilkin: You love that `Array` coercion... My thought was that we'd add lower level tests to ensure an enumerable was return from that method once we know more about what it's meant to do

gjvis:
I think this is a solid start.

There's a question of feature flags before we merge this into master?

I'm personally comfortable with this going live without a feature flag as long as we're very careful about defensive coding (e.g. the Array coercion), so that we don't create situations where we're throwing errors if anyone happens to hit these pages (because logs, and because errors are potentially alarming)

What do others think?

app/controllers/locations_controller.rb
gjvis: http://en.wikipedia.org/wiki/Robustness_principle

gjvis: It currently errors if you hit it in a browser, because fragile...

stevenwilkin: I originally had that function return some static data, I have it returning an empty array now until we get some specs in place

gjvis: :+1:

andrewgarner: As we discussed when writing the gherkins, we have no concept of what is 'near' just what it is 'nearest'. I think the API for `Location` should reveal this.

```rb
Locations.nearest_to_postcode
```

features/step_definitions/appointment_locator_steps.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  allow(Locations).to receive(:near_postcode).with(postcode) do
+    1.upto(5).map { |x| Locations::Location.new("cab-#{x}") }
+  end
+
+  Pages::Locations.new.load(postcode: postcode)
+end
+
+Then(/^I should see the (\d+) appointment locations nearest to that postcode$/) do |_number_of_locations|
+  expected_locations = %w(cab-1 cab-2 cab-3 cab-4 cab-5)
```
andrewgarner: As we're ignoring `_number_of_locations` the convention would be to name it `_`, however, why are we ignoring it? At the moment it's a coincidence that `expected_locations` contains the same number of items we've used in our gherkin.

Maybe?

```rb
expected_locations = number_of_locations.times.map { |i| "cab-#{i+1}" }
```

app/controllers/locations_controller.rb
andrewgarner:
```rb
params[:postcode]
```

gjvis: :+1: to both comments

features/step_definitions/appointment_locator_steps.rb
gjvis: or remove the variable completely?

andrewgarner: Looking at this a bit further, maybe it would be better if the mock on `Locations.near_postcode` returned more than X locations. We can then assert that only X are being displayed.

andrewgarner: Also, 5 isn't dynamic so we shouldn't be matching it dynamically. The business rules dictate that we specifically only want to see 5.

stevenwilkin:
Updated to take into consideration feedback, will merge when build passes

PR #194, opened 2015-06-10
Geocode queried postcode

stevenwilkin:
It's one line to call the geocoder... Are we ok with just putting that call in a method or do we want to abstract it somehow?


stevenwilkin:
As discussed with @gjvis, I'm going to flesh this PR out to include all the required collaborators for `Locations.nearest_to_postcode`. These collaborators will be faked in this pass.

PR #196, opened 2015-06-11
Locations search

stevenwilkin:


stevenwilkin:
Updated with new `Locations::SearchResult` and displaying the distance away for each rendered result

gjvis:
:+1:

features/step_definitions/appointment_locator_steps.rb
```diff
@@ -7,12 +8,15 @@
...
   allow(Locations::Search).to receive(:nearest_to).with(
-    locations, lat_lng, limit).and_return(locations.take(limit))
+    locations, lat_lng, limit).and_return(search_results.take(limit))
 
   Pages::Locations.new.load(postcode: postcode)
 end
 
 Then(/^I should see the (\d+) appointment locations nearest to that postcode$/) do |number_of_locations|
   expected_locations = number_of_locations.to_i.times.map { |i| "cab-#{i + 1}" }
+  expected_distances = number_of_locations.to_i.times.map { |i| "#{i}km" }
```
andrewgarner: Why kilometres not miles?

gjvis: Haha, I totally missed that...

stevenwilkin: Good point heh

andrewgarner: We should also make sure the number is rounded and we show the correct number of decimal places.

stevenwilkin:
Updated to use miles in search results

features/step_definitions/appointment_locator_steps.rb
gjvis: How about we leave that for a presenter spec, when we implement an actual view – we don't want to be partially testing things in cucumber at this stage. Sticking with whole numbers for now keeps us moving forward with driving out the implementation.

Everyone ok with that?

andrewgarner: :+1:

PR #197, opened 2015-06-12
Locations repository - read geojson

stevenwilkin:


lib/locations/repository.rb
```diff
@@ -1,7 +1,29 @@
 module Locations
   class Repository
-    def self.all
-      []
+    def initialize(path = Rails.root.join('lib', 'locations', 'locations.json'))
```
gjvis: Pull the default off the Locations module (or some config object in the Locations area) and it'll be easier to specify an alternative path for test data

features/step_definitions/appointment_locator_steps.rb
```diff
@@ -1,11 +1,11 @@
 When(/^I search for appointment locations near to a valid postcode$/) do
-  locations = 1.upto(7).map { |x| Locations::Location.new("cab-#{x}", [1, 0]) }
+  locations = 1.upto(7).map { |x| Locations::Location.new("cab-#{x}", '', '', [1, 0]) }
```
andrewgarner: Maybe time to upgrade Location's initialisation? It would be preferable for address and hours to be optional instead of having to pass in a blank string.

lib/locations/repository.rb
```diff
@@ -1,7 +1,29 @@
 module Locations
   class Repository
-    def self.all
-      []
+    def initialize(path = Rails.application.config.locations_path)
```
gjvis: Rails coupling :scream:

gjvis:
:+1:

stevenwilkin:
Updated to remove Rails config dependency

PR #199, opened 2015-06-15
Use Postcodes.io for geocoding postcodes

stevenwilkin:
With the GMaps geocoder an incomplete postcode like "BT7" still returned a lat/lng which is not the case with this alas.

We're still using `Geocoder` to calculate geographic distances. I've briefly looked at how `RGeo` does that and my head nearly exploded. `Geocoder` is much simpler imho. Any thoughts?


gjvis:
We can call out searching by outcode (the first part of a postcode) as a separate scenario.

:+1:

PR #202, opened 2015-06-16
Styleguide options table

aduggin:


andrewgarner:
Let's merge this into `options-guides`. We've added styles that will impact the current table markup.

PR #205, opened 2015-06-16
Style phone number on iPhone.

jiannis:
Safari on iPhone adds an `<a href=tel:…>` to anything it recognises as a phone number.

cc: @aduggin


aduggin:
:+1:

PR #206, opened 2015-06-16
Use only one Geo library

stevenwilkin:
`RGeo` is now a dependancy so remove use of `Geocoder`.

**NOTE** `rgeo` was originally a dependency of `rgeo-geojson` so I've explicitly added it to the Gemfile


gjvis:
:+1:

PR #208, opened 2015-06-17
Update to Rails 4.2.2

stevenwilkin:
http://weblog.rubyonrails.org/2015/6/16/Rails-3-2-22-4-1-11-and-4-2-2-have-been-released-and-more/


Gemfile.lock
```diff
@@ -270,6 +270,3 @@ DEPENDENCIES
   utf8-cleaner
   vcr
   webmock
-
-BUNDLED WITH
```
andrewgarner: If you update to the latest version of Bundler this will get generated automatically. I think we should keep this tracked or we're going to be constantly excluding these lines.

stevenwilkin: Ah, I saw mention of this on the twitters this morning...

andrewgarner:
:+1:

PR #209, opened 2015-06-17
Update all gems

stevenwilkin:
Catches patch for latest Rack CVE:

http://www.openwall.com/lists/oss-security/2015/06/16/14


gjvis:
:+1:

PR #210, opened 2015-06-17
Break options guide down into multiple guides

andrewgarner:
There are still some outstanding presentational changes that are pending but it would be good to go some feedback on the overall approach.


app/views/components/_journey_nav.html.erb
```diff
@@ -1,46 +1,37 @@
...
-              <a href="#cash" id="sub-nav-cash">Cash your pot in</a>
-            </li>
-            <li class="t-journey-subnav-step">
-              <span class="bullet bullet--mix"></span>
-              <a href="#mix" id="sub-nav-mix">Mix your options</a>
-            </li>
+            <% step.steps.each do |child_step| %>
+              <li class="t-journey-subnav-step<%= child_step.current? ? ' is-current' : '' %>">
+                <span class="bullet bullet--<%= child_step.slug %>"></span>
+                <%= link_to child_step.url do %>
```
gjvis: why `link_to` when we're not using any of the magic ActiveRecord url inference?

app/views/components/_journey_nav.html.erb
```diff
@@ -1,46 +1,37 @@
...
+                  <%= child_step.label %>
+                  <%= content_tag :span, '(current)', class: 'visuallyhidden' if child_step.current? %>
```
gjvis: why `content_tag`?

app/views/guides/show.html.erb
```diff
@@ -2,20 +2,18 @@
...
-      </p>
-
-      <hr class="hr-thin">
-
-      <% cache(@guide.slug) do %>
-        <%= render 'components/journey_nav', journey: @journey, current_step: @guide %>
-      <% end %>
-    </div>
+  <% if journey_navigation.active? %>
+    <% cache(@guide.slug) do %>
```
gjvis: the journey nav being active will also be determined based on the slug, so maybe we can move the caching one line up?

app/views/components/_journey_nav.html.erb
andrewgarner: Fixed with ea6e587.

app/views/components/_journey_nav.html.erb
andrewgarner: Fixed with ea6e587.

app/views/guides/show.html.erb
andrewgarner: Fixed with 3f9a0b2.

spec/view_models/journey_navigation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:baz_guide) { instance_double(Guide, content_type: :govspeak, id: double, slug: 'baz', label: 'Baz') }
+
+  let(:tree) do
+    Tree::TreeNode.from_hash(
+      ['foo', foo_guide] => { ['bar', bar_guide] => { ['baz', baz_guide] => nil } }
+    )
+  end
+
+  describe '.overview' do
+    subject { journey_navigation.overview }
```
gjvis: I think the readability might be improved with a named subject, perhaps `overview_step`, or even just `step`?

spec/view_models/journey_navigation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      let(:current_step) { 'baz' }
+
+      it 'is a current step' do
+        expect(subject.current).to be_truthy
+      end
+    end
+  end
+
+  describe '#steps' do
+    subject { journey_navigation.steps }
```
gjvis: Same here, I think naming the subject `steps` would help readability

spec/view_models/journey_navigation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it 'returns steps for each guide' do
+      expect(subject.first).to be_a(JourneyNavigation::Step)
```
gjvis: we should check all steps if we're calling out "each guide"? also perhaps be explicit that its each "child" guide?

spec/view_models/journey_navigation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it 'has the URL of the guide at the root of the tree' do
+      expect(subject.first.url).to eq('/bar')
+    end
+
+    context 'when the step is the current guide' do
+      let(:current_step) { 'bar' }
+
+      it 'is a current step' do
+        expect(subject.first.current).to be_truthy
```
gjvis: I think this would be clearer if we use `subject.find { |step| step.id == current_step }`

spec/view_models/journey_navigation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+    end
+
+    context 'when a child step is the current guide' do
+      let(:current_step) { 'baz' }
+
+      it 'is a current step' do
+        expect(subject.first.current).to be_truthy
```
gjvis: same here, it'd make the relationship more obvious – i found myself going back to the test setup to understand "what is the first step, how does it related to baz?"

perhaps step names with semantic meaning might help too?

spec/view_models/journey_navigation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it 'has the label of the guide at the root of the tree' do
+      expect(subject.first.label).to eq('Bar')
+    end
+
+    it 'has the URL of the guide at the root of the tree' do
+      expect(subject.first.url).to eq('/bar')
+    end
+
+    context 'when the step is the current guide' do
```
gjvis: what about when it isn't the current guide, falsey is a valid case too

gjvis: (or maybe just add an extra check that only one of `steps` is current?

andrewgarner: Fixed with b7b10e2.

spec/view_models/journey_navigation_spec.rb
andrewgarner: Fixed with b7b10e2.

spec/view_models/journey_navigation_spec.rb
andrewgarner: Fixed with b7b10e2.

spec/view_models/journey_navigation_spec.rb
andrewgarner: Fixed with b7b10e2.

spec/view_models/journey_navigation_spec.rb
andrewgarner: Fixed with b7b10e2.

app/view_models/journey_navigation.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  attr_accessor :journey_tree, :current_step
+
+  def initialize(journey_tree, current_step = nil)
+    self.journey_tree = journey_tree
+    self.current_step = current_step
+  end
+
+  def active?
```
gjvis: This needs tests, and should be memoized. BUT, the other option is moving this responsibility out of this class, perhaps get the controller to ask something for the active journey, and then the view can just check if there is one (rather than if one is active).

andrewgarner: Tests added with 277c98a.

gjvis:
:+1: I think this is good to go (post-rebase, obvs)

PR #212, opened 2015-06-18
Put postcode geocoding into it's own thing

stevenwilkin:
There's future work around outcodes to come


gjvis:
:+1:

improves codebase hygiene

PR #213, opened 2015-06-18
Redirect to appointments page when no postcode supplied in query string

stevenwilkin:


app/controllers/locations_controller.rb
```diff
@@ -2,6 +2,8 @@ class LocationsController < ApplicationController
   layout 'full_width'
 
   def index
+    redirect_to(appointments_path) && return unless params[:postcode].present?
```
gjvis:
```ruby
return redirect_to(appointments_path) unless params[:postcode].present?
```

gjvis: for interest's sake:
```ruby
(redirect_to(appointments_path) and return) unless params[:postcode].present?
```
might work reliably. might.

PR #215, opened 2015-06-19
Search for an invalid postcode

stevenwilkin:
Any opinions on the use of exceptions?

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -7,6 +7,9 @@ class LocationsController < ApplicationController
     @locations = locations.map do |location|
       SearchResultDecorator.new(location)
     end
+
+  rescue Geocoder::InvalidPostcode
+    render :invalid_postcode
   end
 
   def show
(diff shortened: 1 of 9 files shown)


gjvis:
Maybe a bit dangerous to use exceptions for control flow (it's not really an exceptional case), but the code is easy to follow. Your call

app/controllers/locations_controller.rb
```diff
@@ -7,6 +7,9 @@ def index
     @locations = locations.map do |location|
       SearchResultDecorator.new(location)
     end
+
+  rescue Geocoder::InvalidPostcode
+    render :invalid_postcode
```
andrewgarner: Wouldn't we want to show the user an error message not an error page?

andrewgarner:
Agree on flow. If we take the responsibility of postcode validation away from `Geocoder` we don't need to handle exceptions in the way we do. We could just invoke `UKPostcode` directly from the controller or if we think the controller is doing too much wrap those invocations up as another object.

stevenwilkin:
My initial thought was to do the postcode validation in the controller but in the scenario where the user supplies an **invalid postcode** which is actually a **valid outcode** we're planning on looking up the postcode and geocoding that.

As discussed with @gjvis, we're going to ice this PR until we know what the UI is going to look like which will tell us what the API for this code should look like. This should solve the error message versus error page problem as well.

Issue #211, opened 2015-06-18
Our BEM style fails linting


gjvis:
Multiple hyphens currently fail scss linting. We probably need a regex for our specific BEM syntax

andrewgarner:
Fixed with 11ca49e

PR #216, opened 2015-06-22
Configure SCSS-Lint for hyphenated BEM convention

andrewgarner:
Is this a suitable fix for #211?


gjvis:
I think so, @aduggin?

aduggin:
Yup, that will do it

PR #223, opened 2015-06-22
Update making_money_last.md

hvonhaaren:
Added State Pension top up link, old request from DWP

### DIFF
--- a/content/making_money_last.md
+++ b/content/making_money_last.md
@@ -24,6 +24,8 @@ You’ll also need to decide when to take your State Pension once you reach Stat
 
 You can’t take the State Pension early but you can push back when you start receiving it - this is called [deferring the State Pension](https://www.gov.uk/deferring-state-pension/what-you-may-get).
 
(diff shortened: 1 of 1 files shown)


content/making_money_last.md
```diff
@@ -24,6 +24,8 @@ You’ll also need to decide when to take your State Pension once you reach Stat
 
 You can’t take the State Pension early but you can push back when you start receiving it - this is called [deferring the State Pension](https://www.gov.uk/deferring-state-pension/what-you-may-get).
 
+You can also [top up your State Pension](https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/414708/factsheet-top-up.pdf) if your don't get the full amount.
```
gjvis: Typo, "your" should probably be "you". Also don't should probably be don’t?

andrewgarner: Shouldn't we be linking to the GOV.UK publication page not the PDF attachment directly?

The HTML page addresses additional needs like a Welsh version, the ability to order a hard copy etc that I assume are still valid for our users?

PR #227, opened 2015-06-22
Use GOV.UK Universal Analytics cookie text

andrewgarner:
GOV.UK now use Universal Analytics and explain what the Universal Analytics cookies are for. It makes sense for us to adopt the same language.


gjvis:
:+1:

PR #228, opened 2015-06-22
[DO NOT MERGE] Load Locations GeoJSON from locator app

stevenwilkin:
Pairs up with [locator#12](https://github.com/guidance-guarantee-programme/locator/pull/12)


config/initializers/locations.rb
```diff
@@ -1 +1 @@
-Locations.path = Rails.root.join('lib', 'locations', 'locations.json')
+Locations.path = 'http://0.0.0.0:8080/locations.json'
```
andrewgarner: Would we still call this a ‘path’?

stevenwilkin: It still works when it's a path on the local filesystem, which is what happens [with the acceptance features](https://github.com/guidance-guarantee-programme/pension_guidance/blob/28db163d16ed04da16ac0e1b3b9fff945dc6180e/features/support/env.rb#L6).

andrewgarner: It's confusing. I have to know that `Locations::Repository` is using open-uri and therefore I can provide a URL and it will work.

gjvis: I think the quickest thing is to name it SUPER explicitly, something like `Locataions.geojson_path_or_url`, and move on with our lives

stevenwilkin:
Ready to merge @andrewgarner

andrewgarner:
:+1:

PR #239, opened 2015-06-23
Search results drill down

stevenwilkin:
Needs rebased when #238 lands to remove ccae4cc.

Search results:

image

Links through to:

image

### DIFF
--- a/features/pages/location_page.rb
+++ b/features/pages/location_page.rb
@@ -1,0 +1,10 @@
+module Pages
+  class Location < SitePrism::Page
+    set_url '/locations/{id}'
+
+    element :name, '.t-name'
+    element :address, '.t-address'
+    element :phone, '.t-phone'
+    element :hours, '.t-hours'
+  end
+end
--- a/spec/controllers/locations_controller_spec.rb
+++ b/spec/controllers/locations_controller_spec.rb
@@ -1,23 +1,46 @@
-RSpec.describe LocationsController, 'GET index', type: :controller do
-  before do
-    allow(Locations).to receive(:nearest_to_postcode).and_return([])
+RSpec.describe LocationsController, type: :controller do
+  describe 'GET index' do
+    before do
+      allow(Locations).to receive(:nearest_to_postcode).and_return([])
+    end
+
+    specify 'without a postcode param' do
+      get :index
+
+      expect(response).to redirect_to(appointments_path)
+    end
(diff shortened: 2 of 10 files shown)


spec/controllers/locations_controller_spec.rb
```diff
@@ -1,23 +1,48 @@
...
+
+    before do
+      allow(Locations).to receive(:find).with(invalid_id)
+      allow(Locations).to receive(:find).with(valid_id).and_return(location)
+    end
+
+    specify 'with an invalid id' do
+      get :show, id: invalid_id
+
+      expect(response).to redirect_to(appointments_path)
```
gjvis: why not simply 404? I wonder if there are any implications to doing this in terms of caching etc. I also wonder what others on the team think in terms of usability

gjvis:
Looks great, one small question around invalid id behaviour...

spec/controllers/locations_controller_spec.rb
stevenwilkin: Good spot, the error scenarios in `LocationsController#index` redirect and I'd blindly done the same thing here

stevenwilkin:
Rebased. Renders standard 404 when requesting an invalid location id

gjvis:
Looks good :+1:

features/pages/location_page.rb
```diff
@@ -0,0 +1,10 @@
+module Pages
+  class Location < SitePrism::Page
+    set_url '/locations/{id}'
```
andrewgarner: Is this a conscious decision to include the resource type in the URL?

We haven't followed that convention with guides for example.

stevenwilkin: No. This is building on what is [already in place with `Pages::Locations`](https://github.com/guidance-guarantee-programme/pension_guidance/blob/37fae3cf14fdeb87098c8fb270efe1e0704c0ab4/features/pages/locations_page.rb#L3).

gjvis: It is a conscious decision yes. It was defined as a `resources` in the routes file to facilitate exactly this.

stevenwilkin:
Reminder to remove a90ed906b8cfb2851a282d3f6635c00003f76257 and d23fb3be8f0cc9114ac88750a72447e7c1ecaecb when the relevant PRs are merged.

PR #241, opened 2015-06-24
@wip location handles own booking

stevenwilkin:


features/step_definitions/appointment_locator_steps.rb
```diff
@@ -37,3 +41,21 @@
...
+                when 'its name'
+                  :name
+                when 'its address'
+                  :address
+                when 'its opening hours'
+                  :hours
+                when 'its Pension Wise booking phone number'
+                  :phone
+                end
+    expect(location.public_send(attribute)).to be_visible
```
gjvis: If we can assert the actual values then we get a slightly more complete edge-to-edge test. Perhaps worth it?

PR #242, opened 2015-06-24
@wip bookmark search results

stevenwilkin:
Have discussed the possibility of deleting this scenario with @gjvis but here you go.


gjvis:
This is superceded by #240 

We probably need to move the VCR into an [around hook](https://github.com/cucumber/cucumber/wiki/Hooks) for this to work simply

stevenwilkin:
This is similar to #240 but it's a different scenario. I've opened #248 concerning VCR cassettes.

PR #245, opened 2015-06-25
Derive a slug-like id from Location name

stevenwilkin:
This needs more discussion but it's ripped out of #239 and can potentially go straight onto `master`.

### DIFF
--- a/lib/locations/location.rb
+++ b/lib/locations/location.rb
@@ -1,3 +1,7 @@
 module Locations
-  Location = Struct.new(:name, :address, :phone, :hours, :lat_lng)
+  Location = Struct.new(:name, :address, :phone, :hours, :lat_lng) do
+    def id
+      name.downcase.gsub(/\W/, '-').gsub(/-{2,}/, '-')
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


gjvis:
perhaps we should use the lat and long in the id as well, to enforce uniqueness? (if two locations have the same lat, long, and name, then I think we can assume they're effectively the same place?)

stevenwilkin:
We have 3 duplicate titles in the CitA data:
```bash
$ grep title *json | sed 's/\",$//' | sed 's/.*\"//' | sort > 1.txt
$ grep title *json | sed 's/\",$//' | sed 's/.*\"//' | sort | uniq > 2.txt
$ diff 1.txt 2.txt
33d32
< Bedford & District CAB
136d134
< Citizens Advice Shropshire - Shrewsbury
249d246
< Hammersmith and Fulham Citizens Advice Bureau
$
```
`Bedford & District CAB` appears to be a straight-forward duplicate but the other 2 have different addresses.

I feel confident in merging this now and dealing with the data problems later as we already have a card in the backlog to cover it.

PR #252, opened 2015-06-29
Don't redirect when no postcode supplied

stevenwilkin:
This view will need the search form added to it but will mean people can browse to `/locations`:

image


andrewgarner:
I'm wondering if `/locations/search` would be better?

stevenwilkin:
This PR makes our search behave like [gov.uk/search](https://www.gov.uk/search) which I think is acceptable:

image

gjvis:
I think this is good. If we went with `/locations/search` then we still need to decide what to show at `/locations`

PR #256, opened 2015-06-30
Use the GeoJSON Feature ID to identify locations

andrewgarner:
Generating a slug for each location based on its title is too brittle. We know location titles will change. If that happens previously addressable locations will no longer be addressable.

The GeoJSON specification supports an ID attribute for each location. By using that we can be continue to address a location if its title changes.


gjvis:
:+1:

PR #258, opened 2015-06-30
Caching Locations reader

stevenwilkin:
We're still instantiating the repository for each operation, but:

* JSON pulled from Locator is stored in Redis and expires after an hour
* JSON from local filesystem is read every time


app.json
```diff
@@ -9,6 +9,7 @@
   },
   "addons": [
     "fastly:fast",
-    "logentries:starter"
+    "logentries:starter",
+    "redistogo"
```
andrewgarner: Heroku Redis is now generally available (`heroku-redis`). I would favour using it over a third party add-on.

lib/locations/reader.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def json_from_url?
+      path.to_s =~ /^http/
+    end
+
+    def cached_json
+      json = redis.get('locations')
+      unless json
+        json = read_json(path)
+        redis.setex('locations', 60 * 60, json)
```
andrewgarner: Should we be determining the cache duration or observing the headers sent by the HTTP service?

app.json
stevenwilkin: Sweet, I didn't know this

gjvis: Yeah, good shout

lib/locations/reader.rb
gjvis: At the very least lets make this configurable by the environment?

spec/lib/locations/reader_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'fakeredis'
+
+RSpec.describe Locations::Reader, '#call' do
+  let(:now) { Time.zone.local(2015, 6, 30, 14) }
+  let(:key) { 'locations' }
+  let(:json) { 'json' }
+  let(:path) { '/path/to/locations.json' }
+  let(:redis) { Redis.new }
+  let(:json_in_redis) { redis.get(key) }
```
gjvis: `let` is for memoized context, I fear this will cause future sadness...

lib/locations/reader.rb
stevenwilkin: The Go app isn't setting any headers currently and I'm not sure if we're using CloudFlare for this either.

The net effect of this code is to merely reduce the need for an outgoing HTTP request every time the repository is instantiated.

As has been touched on briefly, this could be extended into a cron job which pulls down the locations GeoJSON every hour which would allow the guidance app to just look in Redis for the data.

spec/lib/locations/reader_spec.rb
stevenwilkin: The Redis db is cleared before each spec is run

gjvis: Same length, very obviously not test context (even though it is a bit ugly):
```ruby
def json_in_redis; redis.get(key) end
```

stevenwilkin:
Updated to remove abuses of RSpec syntax

andrewgarner:
:+1:

andrewgarner:
This code is dependent on Redis being present in the environment. We need to upgrade our environments before merging.

stevenwilkin:
Rebased. Needs `app.json` updated with whichever Heroku Redis addon we go for.

spec/lib/locations/reader_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  def expiry_in_seconds
+    redis.ttl(key)
+  end
+
+  subject(:read) { reader.call }
+
+  before do
+    Timecop.freeze(now)
+    redis.flushdb
+    ENV['LOCATIONS_TTL'] = locations_ttl.to_s
```
andrewgarner: Should we be modifying the environment like this and not stubbing it?

lib/locations/reader.rb
```diff
@@ -0,0 +1,43 @@
...
+      json = redis.get('locations')
+      unless json
+        json = read_json(path)
+        redis.setex('locations', locations_ttl, json)
+      end
+      json
+    end
+
+    def locations_ttl
+      ENV['LOCATIONS_TTL'].to_i
```
andrewgarner: What happens if this isn't set? Should there be a sensible default?

spec/lib/locations/reader_spec.rb
gjvis: Agreed. Perhaps passing it in on the constructor (with a default set off the ENV) would be cleaner?

gjvis:
:+1:

PR #259, opened 2015-07-01
Add `booking_centre` field to `Location`

stevenwilkin:


gjvis:
We sure that this is how we want to do this? We could reference other booking centres ID?

stevenwilkin:
Currently the data we have uses the title of the booking centre. We could update it to use IDs certainly, and all that would need to be done here is update the fixture and spec.

gjvis:
Yeah sure, just wondering whether now is the time to enhance the domain model?

stevenwilkin:
Makes sense, as long as we update the data accordingly.

stevenwilkin:
Updated this to use IDs

lib/locations/location.rb
```diff
@@ -1,3 +1,3 @@
 module Locations
-  Location = Struct.new(:id, :name, :address, :phone, :hours, :lat_lng)
+  Location = Struct.new(:id, :name, :address, :booking_centre, :phone, :hours, :lat_lng)
```
andrewgarner: Given booking centre is just another location, wouldn't `booking_location` be more revealing?

spec/fixtures/locations.json
```diff
@@ -14,6 +14,7 @@
       "properties": {
         "title": "Belfast Citizens Advice Bureau",
         "address": "Merrion Business Centre, 58 Howard St, Belfast, BT1 6PJ",
+        "booking_centre": "928543c7-e8f0-45b2-a300-c2f996b007d9",
```
andrewgarner: `booking_location_id`?

gjvis: yeah, probably worth calling out that its the ID

stevenwilkin: The GeoJSON coming out of `Locator` has `booking_centre`

gjvis: maybe that should change too?

andrewgarner: It exposes `"booking_centre": "Greendale Citizens Advice"` so this change is coupled to the GeoJSON changing too. Hanging off `booking_location_id` lets us retire `booking_centre` down the line.

stevenwilkin: It would be a good thing to do though requires a fair bit of rejigging of the JS app.

I've opened [locator#20](https://github.com/guidance-guarantee-programme/locator/issues/20) to track this.

Stick with `booking_centre` until this is done?

stevenwilkin: Took me longer than anticipated to write that last comment ;)

So, the plan is to start using `booking_location_id` in this app *now* while the GeoJSON still contains `booking_centre` which will be retired in due course.

stevenwilkin:
Updated to use `booking_location_id `

andrewgarner:
:+1:

gjvis:
:+1:

PR #260, opened 2015-07-01
Render a new search page if no postcode given

andrewgarner:


app/views/locations/invalid_postcode.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for(:page_title, "Face-to-face locations near you - Pension Wise") %>
```
stevenwilkin: Should the title be _Find a face-to-face appointment location..._ like in `search.html.erb`? The `h1`s are the same for both these views.

andrewgarner: Yes - i'll fix.

andrewgarner:
Title amended @stevenwilkin

stevenwilkin:
:+1:

PR #261, opened 2015-07-02
Attempt to fix SASS import bug

andrewgarner:
Upgrades us to the latest rails/sass-rails version. This includes a fix for caching of globbed import statements which is a problem we are seeing when deploying.


gjvis:
:+1:

PR #265, opened 2015-07-03
Link to relevant appointment content from location profile

andrewgarner:
Linking to the “What you’ll get from the appointment” heading is more useful to the user since all content before that on the page is about phone booking and who can book.

cc/ @alexkouz @aduggin


gjvis:
:+1:

PR #266, opened 2015-07-03
Delivery location with different booking location

stevenwilkin:
Delivery locations which have a booking location display the opening hours and phone number of the booking location along with the name of the booking location:

image


andrewgarner:
:+1:

PR #268, opened 2015-07-03
Return cache headers for location pages

andrewgarner:


config/initializers/locations.rb
```diff
@@ -1 +1,5 @@
-Locations.geo_json_path_or_url = ENV['LOCATIONS_URL'] || 'https://locator.pensionwise.gov.uk/locations.json'
+Locations.geo_json_path_or_url = if Rails.env.test?
+                                   Rails.root.join('features', 'fixtures', 'locations.json')
```
stevenwilkin: This is set in [env.rb](https://github.com/guidance-guarantee-programme/pension_guidance/blob/6a3173d00c69c97feab8dfe95e210ab1ac12d128/features/support/env.rb#L6)

andrewgarner: That's only applicable for the acceptance tests.

stevenwilkin: We should set it only once then? Removing it from `env.rb` would probably be the thing to do

andrewgarner: :+1:

andrewgarner: @stevenwilkin - done and rebased.

PR #294, opened 2015-08-03
Add address of location to search results

andrewgarner:
screen shot 2015-08-03 at 09 56 36


stevenwilkin:
Looks good, just needs the cukes fixed

PR #295, opened 2015-08-03
Explain the role of Pension Wise / Citizens Advice when searching for face-to-face locations

andrewgarner:


stevenwilkin:
This simplifies things, I like it :+1: 

The ['booking location name' case branch](https://github.com/guidance-guarantee-programme/pension_guidance/blob/572b0e7b650d4df1f30fa8f5a7e6eb6fcd3fbb4d/features/step_definitions/appointment_locator_steps.rb#L86-L87) should be removed.

We should be able to remove [`LocationDecorator#booking_location`](https://github.com/guidance-guarantee-programme/pension_guidance/blob/8889d7f5176d8fe3c2642091fa7a6184354df117/app/decorators/location_decorator.rb#L15-L17) as well if the booking location is not exposed in the view?

PR #296, opened 2015-08-03
Surface information about a location to the dataLayer

andrewgarner:
Surfaces data about locations to the Google Tag Manager dataLayer.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,6 +45,7 @@ group :test do
   gem 'cucumber-rails', require: false
   gem 'fakeredis'
   gem 'launchy'
+  gem 'poltergeist'
   gem 'scss-lint', '~> 0.30'
   gem 'shoulda-matchers'
   gem 'site_prism'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,6 +52,7 @@ GEM
       rack (>= 1.0.0)
       rack-test (>= 0.5.4)
       xpath (~> 2.0)
+    cliver (0.3.2)
     coderay (1.1.0)
     crack (0.4.2)
       safe_yaml (~> 1.0.0)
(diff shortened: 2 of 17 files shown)


features/step_definitions/location_steps.rb
```diff
@@ -17,7 +17,7 @@
   location.name.click
 end
 
-When(/^I view the details of an appointment location that handles its own booking$/) do
+When(/^(?:I view|a user views) the details of an appointment location(?: that handles its own booking)?$/) do
```
stevenwilkin: Perhaps define a separate `When` instead of introducing complexity into the regex here?

stevenwilkin:
Any reason for testing this with Cucumber?

This feature isn't adding any value to the end user and my concern is that it's bloating the set of acceptance criteria for the app

features/support/matchers/have_data_layer_information.rb
```diff
@@ -0,0 +1,19 @@
...
+  def wait_until(timeout_in_seconds = Capybara.default_wait_time)
+    Timeout.timeout(timeout_in_seconds) do
+      sleep(0.1) until (value = yield)
+      value
+    end
+  end
+
+  match do |page|
+    wait_until do
+      @gaq = Array(page.evaluate_script('window.dataLayer'))
```
stevenwilkin: Does this need to be an instance variable?

stevenwilkin:
Can the location id not be deduced from the page url in GA?

andrewgarner:
I started off with a request spec, but, Capybara is no longer supported in
request specs as of Capybara 2.0.0. The recommended way to use Capybara is
with feature specs.

Given we have Cucumber features I decided to follow the same path. I also think there is value in documenting the need of an analyst with gherkins. I'd argue they are an end-user of this application.

stevenwilkin:
Using Gherkin to assert that a specific piece of JavaScript has been run just seems wrong to me. If this *has* to be done I'd much rather see it done with RSpec.

andrewgarner:
cc/ @guidance-guarantee-programme/developers - now testing with RSpec.

stevenwilkin:
03485a74b74fd690449136fa0d06e97bcca5f6c1 refactors a lot of the postcode stuff in the controller along with the dataLayer piece. Would be easier to follow if this was 2 commits.

stevenwilkin:
VCR cassettes could be better in their own commit

andrewgarner:
In regards to committing the cassettes separately – shouldn't aspire for each commit to be green?

stevenwilkin:
It's an ideological issues I guess, I just find it easier to review code when changes to cassettes are segregated.

PR #304, opened 2015-08-06
Add search form to booking page and make forms consistent

aduggin:
This is a branch of locations/journey


aduggin:
This is a branch of locations/journey

PR #312, opened 2015-08-12
Improve 'Elsewhere' links on location pages

aduggin:


app/views/locations/show.html.erb
```diff
@@ -45,3 +45,23 @@
 <p>
   You can also book in person by visiting any Citizens Advice bureau.
 </p>
+
+<% content_for :sidebar do %>
```
stevenwilkin: Any reason for moving the `sidebar` content block out of `layouts/locations.html.erb` and into this view?

app/controllers/guides_controller.rb
```diff
@@ -48,7 +48,12 @@ def decorate(guide_or_collection)
   end
 
   def non_journey_related_guides
-    @related_guides ||= guide_repository.find_all(*NON_JOURNEY_RELATED_GUIDE_IDS)
+    @related_guides ||= begin
+      ids = *NON_JOURNEY_RELATED_GUIDE_IDS
```
stevenwilkin: Can just use ``ids = NON_JOURNEY_RELATED_GUIDE_IDS``, don't need the splat.

app/controllers/guides_controller.rb
```diff
@@ -48,7 +48,12 @@ def decorate(guide_or_collection)
...
+      ids.reject! { |id| id == 'appointments' } if params[:id] == 'book'
```
stevenwilkin: There's no tests for this.

app/views/locations/show.html.erb
andrewgarner: Because it is only required when viewing a location, not for all other location pages.

stevenwilkin: Fair enough!

PR #314, opened 2015-08-18
Tell IE11+ there is no "browserconfig.xml"

andrewgarner:
If a webpage does not specify a browser configuration file, IE11+ automatically looks for "browserconfig.xml" in the root directory of the server. To prevent this, use an "msapplication-config" header with the content attribute to "none". See https://msdn.microsoft.com/en-us/library/dn320426(v=vs.85).aspx


stevenwilkin:
IE11 WTF? :+1:

PR #320, opened 2015-08-26
Refactor LocationDecorator construction

stevenwilkin:
This extracts logic from the controller and decorator into more easily testable locations. Introduces even more small parts to the system but should hopefully be easier to understand and modify in the future.


spec/services/create_location_decorator_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  subject(:decorator) { create_decorator }
+
+  before do
+    allow(Locations).to receive(:find).with(booking_location_id).and_return(booking_location)
+    allow(Locations).to receive(:nearest_to_postcode).with(postcode, limit: search_limit).and_return(nearest_locations)
+    allow(LocationSearchContext).to receive(:build).with(nearest_locations, location).and_return(search_context)
+    allow(Switchboard).to receive(:lookup).with(location_id).and_return(twilio_number)
+    expect(LocationDecorator).to receive(:new).with(
```
andrewgarner: Why are we setting expectations in a before step?

spec/services/create_location_decorator_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    let(:booking_location_id) { nil }
+    let(:booking_location) { nil }
+
+    it 'does not retrieve the booking location' do
+      expect(Locations).to_not receive(:find)
+      create_decorator
+    end
+  end
+
+  context 'when the cab locator raises an exception' do
```
andrewgarner: `cab locator` isn't a domain term. How about `when finding locations nearest to a postcode raises an exception`?

spec/services/create_location_decorator_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def create_decorator
+    described_class.build(location: location,
+                          postcode: postcode,
+                          search_limit: search_limit)
+  end
+
+  subject(:decorator) { create_decorator }
+
+  before do
```
andrewgarner: I'm on the fence as to whether all this stubbing is good/bad given this is a service object. It feels like we're just testing our stubs. What do you think?

app/services/create_location_decorator.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLocationDecorator
```
andrewgarner: Having a service object to create a decorated object doesn't quite feel right.

I'm wondering whether the service could return a value object containing `location`, `booking_location`, `search_context` and `twilio_number` then the controller can take the responsibility of decorating for the view it is trying to render?

spec/services/create_location_decorator_spec.rb
stevenwilkin: You're right, `cab locator` isn't an actual part of our domain, it just keeps creeping in :/

spec/services/create_location_decorator_spec.rb
stevenwilkin: Seeing as this is an extraction of existing functionality the intention was just to ensure the correct messages were being passed around, ie that the correct collaborators were being involved in the correct way. Each of these components is tested separately and *mostly* covered with higher-level acceptance criteria.

spec/services/create_location_decorator_spec.rb
stevenwilkin: Yeah, I was wanting to ensure the decorator was actually constructed when I was building this thing but the assertion in the `it { ... }` below would imply that.

app/services/create_location_decorator.rb
stevenwilkin: The problem as I see it is that in order to present a `Location` to the view we now have to decorate it along with **three** other collaborating objects.

This service is just a factory method but I do like having just one thing to pass all the parameters to and getting back the object for the view to present. Moving the logic out of the controller I think is a good move especially seeing how we have agreed to stick with Rubocop's guidelines when it comes to method length and complexity.

Having one object construct a decorated object is not inherently bad I think; we have a measure of complexity to deal with and the priority should be to make that as easy to follow as possible for our future selves.

If we think that having an object to coordinate things is a good idea then perhaps we could go a step further and have it in charge of all interactions within the controller action, say something like `FindAndDecorateLocation`.

I see little value in having something that returns a list of values only then for those values to be piped straight into a constructor.

Probably best to discuss this face-to-face tomorrow!

benbarnett:
Should this still be open?

stevenwilkin:
Nobody cared enough about it :cry:

PR #321, opened 2015-08-28
Revert "Remove rake task fix for jasmine"

stevenwilkin:
The fix introduced in 0372f76226c1171b88b93e36f372b5764c901993 was reverted in c9bcc895407e7c8180094af410e5b9a23378250e, however running `rake spec:javascript` locally now breaks. Revert it.

Any opinions @andrewgarner ?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,4 +17,18 @@ begin
 rescue LoadError
 end
 
+Rake::Task['spec:javascript'].clear
+
+namespace :spec do
+  desc 'Run the code examples in spec/javascripts with PhantomJS'
+  task javascript: [:environment] do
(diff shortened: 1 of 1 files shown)


andrewgarner:
Annoyingly my revert commit doesn't allude to why this "was not needed"  😡

I seem to remember it causing a problem with Heroku deployments but I can't find the evidence in the build history.

I've been setting `RAILS_ENV` to `test` when I run tests locally to ensure a successful run.

If you can get it to deploy then :+1:

stevenwilkin:
I don't run `rake` locally so I didn't notice and it's been building on Travis ok. I'll merge it and see what happens...

PR #324, opened 2015-09-01
Media Player

andrewgarner:
Adds a video with captions and transcript and progress data published to the GTM data layer

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,6 +1,7 @@
 //= require jquery/dist/jquery.min.js
 //= require click-tracker.js
 //= require media-player.js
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/application.js
```diff
@@ -1,4 +1,7 @@
 //= require jquery/dist/jquery.min.js
 //= require click-tracker.js
+//= require media-player.js
+//= require govuk_elements/public/javascripts/vendor/details.polyfill.js
```
andrewgarner: Maybe we should add `vendor/assets/bower_components/govuk_elements/public/javascripts/vendor/` to our asset path then we can just require `details.polyfill.js`?

PR #325, opened 2015-09-01
Use Faraday rather than Excon for Switchboard lookups

stevenwilkin:
We're using Faraday for Google Custom Search so be consistent


lib/switchboard.rb
```diff
@@ -20,5 +14,13 @@ def lookup(id)
     def base_url
       ENV['SWITCHBOARD_BASE_URL']
     end
+
+    def connection
```
andrewgarner: Why not use https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/lib/http_connection_factory.rb ?

stevenwilkin: I looked at that but didn't think of using it, good call!

stevenwilkin:
Updated to use `HTTPConnectionFactory `

lib/switchboard.rb
andrewgarner: The approach we've taken with Search is to pull the connection from a registry - https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/config/initializers/search.rb#L1-L2

andrewgarner:
How about using the registry pattern as we have for Search? - https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/config/initializers/search.rb#L1-L2

stevenwilkin:
This is complicated a bit by using the presence of the environment variable as a feature toggle

andrewgarner:
Where do we do that?

stevenwilkin:
Just rebased this in light of [this recent bugsnag](https://bugsnag.com/hm-treasury-1/pension-guidance/errors/56129c2902c5eb75212ef2a3).

The presence of the env var is used as a toggle [here](https://github.com/guidance-guarantee-programme/pension_guidance/blob/use-faraday-for-switchboard/lib/switchboard.rb#L4) @andrewgarner.

stevenwilkin:
Updated to use the registry

andrewgarner:
:+1:

Issue #322, opened 2015-09-01
Postcode.io timeouts


stevenwilkin:
Lookups to [Postcodes.io occasionally timeout](https://bugsnag.com/hm-treasury-1/pension-guidance/errors/55e522b188f8ad53515e9b10) in `lib/geocode.rb`.

PR #331, opened 2015-09-22
Handle Postcodes::IO lookup timeouts

stevenwilkin:
Addresses #322 

The specific exception we've seen raised in production is `Excon::Errors::Timeout` but this covers anything being raised.

Too crude @andrewgarner ? Are we ok with rendering the _xxx is not a valid postcode_ page when this occurs?

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -14,6 +14,8 @@ class LocationsController < ApplicationController
       end
     rescue Geocoder::InvalidPostcode
       render :invalid_postcode
+    rescue Geocoder::FailedLookup
+      render :failed_lookup
     end
   end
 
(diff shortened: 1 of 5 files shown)


andrewgarner:
I think it would be better for the user if we could show a "Out fault, please try again" page?

stevenwilkin:
Do we want to use the generic 500 error page then or just craft one specifically for this scenario then?

andrewgarner:
I'd say create a new view that is similar to the invalid postcode view but with different messaging.

stevenwilkin:
Seems legit

stevenwilkin:
Updated to render this on failure:

image

Is the content too simplistic?

andrewgarner:
@alexkouz will have an opinion.

I think it would make sense to render the form again for them so that they can try again. We know what they're trying to do and we should expect them to have to use the back button.

stevenwilkin:
image

Well?

alexkouz:
Do we not need to specify what went wrong?

stevenwilkin:
The errors we've had in production have been due to a request to an external service timing out. That type of detail would be meaningless for users though?

alexkouz:
Okay, let's leave as per your suggested text.

PR #333, opened 2015-09-25
[DO NOT MERGE] Leave pot untouched calculator prototype

stevenwilkin:
**NOTE**: the JavaScript and controller code are untested and the presentation and copy are rough and ready.

# JS

image

# Non-JS fallback

image

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,6 @@
 //= require jquery/dist/jquery.min.js
 //= require click-tracker.js
+//= require leave-pot-untouched-calculator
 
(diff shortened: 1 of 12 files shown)


stevenwilkin:
Updated to use a shared partial to render the _Your pot would be worth..._ text for JS and non-JS scenarios.

This text has also been moved to after the form to be consistent with previous efforts.

stevenwilkin:
Updated to take into consideration the feedback from lab testing, will need validation added in line with #356 when that gets agreed upon.

stevenwilkin:
Rebased

PR #335, opened 2015-09-29
Don't depend on system for DNS resolution

stevenwilkin:
We've had [a few failed DNS lookups](https://bugsnag.com/hm-treasury-1/pension-guidance/errors/56096ff402c5eb752121b3d6) as described [in this post](http://blog.gregburek.com/2015/02/22/dns-eglibc-and-resolv-replace-on-heroku/).

We can give this a spin for a while and see if the problem disappears.


andrewgarner:
:+1:

PR #340, opened 2015-10-13
Handle the location not being present in the nearest locations

stevenwilkin:
Shouldn't happen but handle it when it does happen, see [this bugsnag I created](https://bugsnag.com/hm-treasury-1/pension-guidance/errors/561ccf3502c5eb75213edadd) whilst checking something else out.


stevenwilkin:
Rebased

PR #343, opened 2015-10-19
Bounce browserconfig requests

stevenwilkin:
I don't like the `require` in `config/application.rb` but there you go.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -15,6 +15,8 @@ module PensionGuidance
     config.cache_max_age = ENV['CACHE_MAX_AGE'] || 10.seconds
 
     config.middleware.use Rack::BounceFavicon
+    require 'bounce_browserconfig'
+    config.middleware.use BounceBrowserconfig
 
     config.mount_javascript_test_routes = false
   end
--- a/lib/bounce_browserconfig.rb
+++ b/lib/bounce_browserconfig.rb
@@ -1,0 +1,13 @@
+class BounceBrowserconfig
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env['PATH_INFO'] == '/browserconfig.xml'
+      [404, { 'Content-Type' => 'text/html', 'Content-Length' => '0' }, []]
+    else
+      @app.call(env)
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


stevenwilkin:
I've disabled the alert in LogEntries for 404s caused by `/browserconfig.xml` also.

config/application.rb
```diff
@@ -15,6 +15,8 @@ class Application < Rails::Application
     config.cache_max_age = ENV['CACHE_MAX_AGE'] || 10.seconds
 
     config.middleware.use Rack::BounceFavicon
+    require 'bounce_browserconfig'
```
andrewgarner: Why don't you like it? Because of it's placement? Can it not go at the top of the file?

stevenwilkin: It can't go at the top of the file as `lib` isn't on the `$LOAD_PATH` yet :/

andrewgarner: But it could go at line 6?

config/application.rb
```diff
@@ -15,6 +15,8 @@ class Application < Rails::Application
...
+    config.middleware.use BounceBrowserconfig
```
andrewgarner: Do you think we should add this to a `Rack` namespace?

config/application.rb
stevenwilkin: Not without pushing `lib` onto `$LOAD_PATH` first, which is what happens on line 13 with `config.autoload_paths << Rails.root.join('lib')`

config/application.rb
stevenwilkin: I had considered it but decided against it purely as a gut reaction to this being a part of our own codebase. Can be changed if you feel strongly enough about it.

config/application.rb
andrewgarner: Ahh OK yeh. You could use `config.middleware.insert_before` from an initialiser?

stevenwilkin: It could be a possibility but `config/application.rb` is the only place in the codebase where we;re explicitly pulling middleware.

stevenwilkin:
Rebased and the build is green.

andrewgarner:
:+1:

PR #344, opened 2015-10-19
Disable `Switchboard` lookups after testing

stevenwilkin:
The connection was lingering around and causing tests hitting `LocationsController#show` to fail whenever they were ran after.

Failure demonstrated with:
```
rspec ./spec/lib/switchboard_spec.rb ./spec/controllers/locations_journey_breadcrumbs_spec.rb:48 --seed 64399
```


andrewgarner:
:+1:

PR #347, opened 2015-11-02
[DO NOT MERGE] Take whole pot calculator

stevenwilkin:
This basically a reworking of [prototype-whole-pot-calculator](https://github.com/guidance-guarantee-programme/pension_guidance/compare/prototype-whole-pot-calculator) but with the logic implemented in Ruby and incorporating the feedback from lab testing.

It follows the style of progressive enhancement introduced with [leave-pot-untouched-calculator](https://github.com/guidance-guarantee-programme/pension_guidance/compare/leave-pot-untouched-calculator) and appears thus on `/take-whole-pot`:
image

The computed result looks like:
image

The look and feel is my best effort at reproducing the user-researchers' design and no doubt fails to live up to @aduggin's legacy so will be good to get @benbarnett's opinion when he comes onboard.

### DIFF
--- a/app/views/take_whole_pot_calculator/_results.html.erb
+++ b/app/views/take_whole_pot_calculator/_results.html.erb
@@ -2,11 +2,11 @@
   Based on what you've told us, if you take your whole <%= number_to_currency(@pot) %>
   pension pot in one go, you could receive:
 
-  <strong class="calculator__result__number"><%= number_to_currency(@pot_received) %></strong>
+  <strong class="calculator__result__number t-pot-received"><%= number_to_currency(@pot_received) %></strong>
 
   after paying:
 
-  <strong class="calculator__result__number"><%= number_to_currency(@pot_tax) %> in tax</strong>
+  <strong class="calculator__result__number t-pot-tax"><%= number_to_currency(@pot_tax) %> in tax</strong>
 </p>
 
 <h2>Things to note</h2>
--- a/app/views/take_whole_pot_calculator/_results.html.erb
+++ b/app/views/take_whole_pot_calculator/_results.html.erb
@@ -1,0 +1,23 @@
+<p>
+  Based on what you've told us, if you take your whole <%= number_to_currency(@pot) %>
+  pension pot in one go, you could receive:
+
+  <strong class="calculator__result__number"><%= number_to_currency(@pot_received) %></strong>
+
+  after paying:
+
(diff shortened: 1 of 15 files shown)


stevenwilkin:
Good to go @andrewgarner ? Just needs rebased

app/views/take_whole_pot_calculator/_results.html.erb
```diff
@@ -0,0 +1,23 @@
+<p>
+  Based on what you've told us, if you take your whole <%= number_to_currency(@pot, unit: '£') %>
```
andrewgarner: How about introducing an `en-GB` locale so that we don’t have to specify the currency unit each time?

app/views/take_whole_pot_calculator/show.html.erb
```diff
@@ -0,0 +1,40 @@
+<h1>Taking your whole pot in one go</h1>
+
+<div class="calculator">
+  <h2>Estimate how much you would get</h2>
+
+  <form action="/take-whole-pot/results" method="post">
```
andrewgarner: I'm not sure it makes sense to use `POST` given we're not creating a resource.

If we used `GET` we're saying "give me this resource, here's some additional context in the form of params".

andrewgarner:
We’re missing an acceptance test (that uses a JS enabled browser).

benbarnett:
Couple of general questions:

* where will the inbound links to the calculator be?
* has the URL been discussed? Perhaps makes sense to live in `/calculators/` or similar

andrewgarner:
My understanding is that it would live within the [Taking your whole pot in one go](https://www.pensionwise.gov.uk/take-whole-pot) guide. It wouldn’t be accessible as a stand-alone calculator.

app/views/take_whole_pot_calculator/show.html.erb
stevenwilkin: My original thought was that we were "creating" a calculation, but what you say has merit: we're not creating anything.

stevenwilkin:
Basic Capybara test added

stevenwilkin:
First pass at running a spec *with* and *without* JavaScript in 52844ba5eaa6178691606fb75d4e4d72c8b4e625

stevenwilkin:
Rebased

stevenwilkin:
Superseded by #359

PR #356, opened 2015-11-09
[DO NOT MERGE] Take whole pot calculator - server-side validation

stevenwilkin:
A clunky first pass at performing side-side validation of calculator input and displaying validation errs per the [GOV.UK guidlines](http://govuk-elements.herokuapp.com/errors):

image

At this stage this doesn't deal with commas in the supplied values and we still need to render values with the correct decimalisation. More helpful validation messages would also be an idea.

### DIFF
--- a/app/forms/take_whole_pot_calculator_form.rb
+++ b/app/forms/take_whole_pot_calculator_form.rb
@@ -3,6 +3,8 @@ require 'active_model'
 class TakeWholePotCalculatorForm
   include ActiveModel::Model
 
+  WEEKS_PER_YEAR = 52
+
   attr_accessor :pot, :income, :pension, :pension_frequency
 
   validates :pot, presence: true, numericality: { greater_than_or_equal_to: 0 }
@@ -20,5 +22,21 @@ class TakeWholePotCalculatorForm
 
   def pension
     Float(@pension.gsub(/,/, '')) rescue @pension
+  end
+
+  def result
+    TakeWholePotCalculator.new(pot, total_income) if valid?
+  end
+
+  private
+
+  def total_income
+    annual_pension = case pension_frequency
+                     when 'weekly'
+                       pension * WEEKS_PER_YEAR
+                     when 'annually'
+                       pension
+                     end
+    income + annual_pension
   end
 end
--- a/app/forms/take_whole_pot_calculator_form.rb
+++ b/app/forms/take_whole_pot_calculator_form.rb
@@ -3,6 +3,7 @@ require 'active_model'
 class TakeWholePotCalculatorForm
   include ActiveModel::Model
 
+  MAXIMUM_STATE_PENSION = 10_400
   WEEKS_PER_YEAR = 52
 
   attr_accessor :pot, :income, :pension, :pension_frequency
@@ -11,6 +12,7 @@ class TakeWholePotCalculatorForm
   validates :income, numericality: true
   validates :pension, numericality: true
   validates :pension_frequency, inclusion: { in: %w(weekly annually) }
+  validate :within_state_pension_limit
 
   def pot
     Float(@pot.gsub(/,/, '')) rescue @pot
@@ -30,13 +32,22 @@ class TakeWholePotCalculatorForm
 
   private
 
+  def within_state_pension_limit
+    errors.add(:pension, :above_limit) if annual_pension > MAXIMUM_STATE_PENSION
+  end
+
+  def annual_pension
+    return 0 unless pension_frequency && pension
+
+    case pension_frequency
+    when 'weekly'
+      pension * WEEKS_PER_YEAR
+    when 'annually'
+      pension
+    end
+  end
+
   def total_income
-    annual_pension = case pension_frequency
-                     when 'weekly'
-                       pension * WEEKS_PER_YEAR
-                     when 'annually'
-                       pension
-                     end
     income + annual_pension
   end
 end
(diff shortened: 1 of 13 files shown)


stevenwilkin:
Updated to accept commas in currency fields

stevenwilkin:
Updated again:
* re-renders the form with the supplied values rather than the type-cast values
* trims trailing zeros from tax and receivable amounts

image

stevenwilkin:
Latest updates in accordance with James' feedback:

* custom error messages
* state pension must be below £10,400/year
* "Back to description" link on results page
* results page is anchored at "Estimate how much you would get" when rendered

app/forms/take_whole_pot_calculator_form.rb
```diff
@@ -0,0 +1,52 @@
...
+  def within_state_pension_limit
+    errors.add(:pension, :above_limit) if annual_pension > MAXIMUM_STATE_PENSION
+  end
+
+  def annual_pension
+    return 0 unless pension_frequency && pension
+
+    case pension_frequency
+    when 'weekly'
+      pension * 52
```
andrewgarner: `52` is a magic number. How about `WEEKS_PER_YEAR`?

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,5 @@
 module ApplicationHelper
+  def format_currency(amount)
+    number_to_currency(amount).sub(/\.00$/, '')
```
andrewgarner: `number_to_currency` takes a `precision` option - http://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_currency

You should be able to drop that in the locale and do away with the customer helper.

stevenwilkin: Using the precision option would cause any non-zero pence values to be dropped?

stevenwilkin:
```ruby
> number_to_currency(100.67, precision: 0)
=> "£101"
```

andrewgarner: Ahh I see. Ignore me then!

config/locales/en.yml
```diff
@@ -36,3 +36,13 @@ en:
...
+
+  activemodel:
+    errors:
+      models:
+        take_whole_pot_calculator_form:
+          blank: 'Using numbers, please enter the value of your %{attribute}'
+          not_a_number: 'Using numbers, please enter the value of your %{attribute}'
+          attributes:
+            pension:
+              above_limit: "You've entered more than the maximum allowable state pension"
```
andrewgarner: `You’ve` not `You've`.

stevenwilkin: Those apostrophes will be the death of me

spec/features/take_whole_pot_calculator_analytics_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    visit '/take-whole-pot'
+
+    fill_in 'pot', with: pot
+    fill_in 'income', with: income
+    fill_in 'pension', with: pension
+    select pension_frequency_label, from: 'pension_frequency'
+    click_on 'Calculate'
+
+    expect(data_layer).to include(
+      'pot' => pot,
```
andrewgarner: This will create invalid Javascript if the user input is a string.

These need to be an integer so that we don't have to clean up the input within GA. So if the user inputs "100,000.00" then in the `dataLayer` we expose `100000`.

spec/features/take_whole_pot_calculator_analytics_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      'income' => income,
+      'pension' => pension,
+      'pension_frequency' => pension_frequency
+    )
```
andrewgarner: I've just spoken to Jim about this – we should also include the `pot_received` and `pot_tax` values. By doing so we won't need to duplicate the calculation logic elsewhere when we do analysis.

app/views/take_whole_pot_calculator/show.html.erb
```diff
@@ -1,3 +1,14 @@
+<% content_for(:data_layer) do %>
+  <script>
+    dataLayer = [{
```
andrewgarner: We should include a variable of `'error': true/false` for additional context.

stevenwilkin:
Updates:
* `valid` field in data exposed to GTM
* second ``Back to description`` link included after form on results page
* robots disallowed from results page

app/views/take_whole_pot_calculator/show.html.erb
```diff
@@ -1,3 +1,19 @@
...
+      'valid': <%= !@result.nil? %>,
```
andrewgarner: Why is this different to `@calculator.errors`?

spec/features/take_whole_pot_calculator_analytics_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      'pension_frequency' => pension_frequency,
+      'valid' => false
+    )
+  end
+
+  scenario 'with valid input' do
+    enter_valid_data
+    perform_calculation
+
+    expect(data_layer).to include(
```
andrewgarner: The values we send to the data layer need to be suitable for Javascript not the human entered values. I've pushed up a change to the feature to expose the problem - 16d6b6a9f6e91e4eb21766de9a7b33f46eeee8fe

stevenwilkin:
Superseded by #359

PR #359, opened 2015-11-16
Take whole pot calculator

stevenwilkin:
This is a combination of #347 and #356 for deployment. These PRs will be closed.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,6 @@
 //= require jquery/dist/jquery.min.js
 //= require click-tracker.js
+//= require take-whole-pot-calculator
 
 PWPG.clickTracker.init();
+PWPG.takeWholePotCalculator.init();
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,6 +1,4 @@
 //= require jquery/dist/jquery.min.js
 //= require click-tracker.js
-//= require take-whole-pot-calculator
 
 PWPG.clickTracker.init();
-PWPG.takeWholePotCalculator.init();
--- a/app/views/take_whole_pot_calculator/show.html.erb
+++ b/app/views/take_whole_pot_calculator/show.html.erb
@@ -1,18 +1,3 @@
-<% content_for(:data_layer) do %>
-  <script>
-    dataLayer = [{
-      'valid': <%= @calculator.valid? %>,
-      'pot': <%= @calculator.pot.to_i %>,
-      'income': <%= @calculator.income.to_i %>,
-      'annual_pension': <%= @calculator.annual_pension.to_i %>
-      <% if @result %> ,
-        'pot_tax': <%= @result.pot_tax.to_i %>,
-        'pot_received': <%= @result.pot_received.to_i %>
-      <% end %>
-    }];
-  </script>
-<% end %>
-
 <a class="link-back" href="/take-whole-pot">Back to description</a>
 
 <h1>Taking your whole pot in one go</h1>
--- a/app/views/take_whole_pot_calculator/show.html.erb
+++ b/app/views/take_whole_pot_calculator/show.html.erb
@@ -1,0 +1,43 @@
+<a class="link-back" href="/take-whole-pot">Back to description</a>
+
+<h1>Taking your whole pot in one go</h1>
+
+<div class="calculator">
+  <h2 id="estimate-how-much">Estimate how much you would get</h2>
+
+  <form action="/take-whole-pot/results#estimate-how-much" method="get">
+    <div class="form-group">
+      <label class="form-label" for="pot">
+        Total value of your pension pot(s)
+      </label>
+      £ <input class="form-control" type="text" id="pot" name="pot" value="<%= @pot %>">
+    </div>
+
+    <div class="form-group">
+      <label class="form-label" for="income">
+        Your total other income for the year (before tax)
+        <span class="form-hint">Earnings and income from any savings or benefits.</span>
+      </label>
+      £ <input class="form-control" type="text" id="income" name="income" value="<%= @income %>">
+    </div>
+
+    <div class="form-group">
+      <label class="form-label" for="pension">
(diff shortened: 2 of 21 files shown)


benbarnett:
This is breakable with negative numbers in the inputs: the validation throws an error correctly but the translation is missing.

screen shot 2015-11-16 at 16 58 18

stevenwilkin:
Good catch @benbarnett !

stevenwilkin:
Addressed with 18da94a5454c33743c1f13c3bc935780fedd2b6f

app/views/take_whole_pot_calculator/show.html.erb
```diff
@@ -0,0 +1,83 @@
+<% content_for(:data_layer) do %>
```
andrewgarner: This should be in `_results.html.erb` so that the behaviour applies when the client has JavaScript?

spec/features/take_whole_pot_calculator_analytics_spec.rb
```diff
@@ -0,0 +1,58 @@
+RSpec.feature 'Take whole pot calculator analytics', type: :feature, js: true do
+  let(:pot) { '100,000' }
+  let(:income) { '7,000' }
+  let(:pension) { '115.95' }
+  let(:pension_invalid) { '500' }
+  let(:pension_frequency) { 'weekly' }
+  let(:pension_frequency_label) { 'Weekly' }
+
+  scenario 'with invalid input' do
```
andrewgarner: We need to apply `with_and_without_javascript` to this.

spec/features/take_whole_pot_calculator_analytics_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    expect(data_layer).to include(
+      'pot' => 100_000,
+      'income' => 7_000,
+      'annual_pension' => 26_000,
+      'valid' => false
+    )
+  end
+
+  scenario 'with valid input' do
```
andrewgarner: We need to apply `with_and_without_javascript` to this.

app/views/take_whole_pot_calculator/show.html.erb
stevenwilkin: My assumption was that we'd push the values into the dataLayer via JS but this could be a better approach

spec/features/take_whole_pot_calculator_analytics_spec.rb
stevenwilkin: A non-JS browser wouldn't be capable of determining the value of `window.dataLayer`?

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,6 @@
+RSpec.describe ApplicationHelper, type: :helper do
+  describe '#format_currency' do
+    specify { expect(helper.format_currency(1_000_000)).to eq('£1,000,000') }
```
andrewgarner: Using the `it "explain behaviour" {}` syntax would improve documentation value of this test:

```rb
describe '#format_currency' do
  subject { helper.format_currency(1_000_000)

  context 'with a whole number' do
    let(:currency) { 1_000_000 }

    it 'removes the decimal digits' do
      expect(subject).to eq('£1,000,000')
    end
  end

  context 'with a decimal number' do
    let(:currency) { 1_000.50 }

    it 'persists the decimal digits' do
      expect(subject).to eq('£1,000.50')
    end
  end
end
```

spec/features/take_whole_pot_calculator_analytics_spec.rb
andrewgarner: Good point, well made.

app/views/take_whole_pot_calculator/show.html.erb
```diff
@@ -0,0 +1,83 @@
...
+  <script>
+    dataLayer = [{
+      'valid': <%= !@result.nil? %>,
```
andrewgarner: How is this different to <%= @calculator.errors.empty? %>

app/views/take_whole_pot_calculator/show.html.erb
stevenwilkin: Moved in eac16ac4340051a3a1678cd9b8894356738d341d. Tried rewriting the commit history but it got messy.

spec/helpers/application_helper_spec.rb
stevenwilkin: That's a lot more descriptive, it's in.

app/views/take_whole_pot_calculator/show.html.erb
stevenwilkin: This is knowledge of [how the calculator works](https://github.com/guidance-guarantee-programme/pension_guidance/blob/take-whole-pot-calculator/app/forms/take_whole_pot_calculator_form.rb#L30)... Could also have `@calculator.valid?`

andrewgarner: That would work yeh.

andrewgarner:
:+1:

Let's showcase and hopefully :ship: first thing tomorrow.

stevenwilkin:
winning

app/assets/javascripts/application.js
```diff
@@ -1,4 +1,5 @@
 //= require jquery/dist/jquery.min.js
 //= require click-tracker.js
+//= require take-whole-pot-calculator
```
andrewgarner: Why are we requiring the calculator JS but then not initialising it?

stevenwilkin: The JS version doesn't have validation. The plan was to get this out and then let @benbarnett loose on it.

andrewgarner: We shouldn't require`take-whole-pot-calculator` and bust the asset cache then.

stevenwilkin: Fair enough, my thoughts were that it would be lower friction to get the code out there and then initialise it in the the next iteration. The require is now gone.

PR #366, opened 2015-11-23
Don't override `window.dataLayer` if it already exists

stevenwilkin:
This is perhaps the cause of the problem Jim has reported. See:

http://www.simoahava.com/gtm-tips/datalayer-declaration-vs-push/


benbarnett:
:+1:

andrewgarner:
:+1:

PR #389, opened 2015-12-21
Site navigation

benbarnett:
Introduces the dropdown navigation. Also removes the `phase banner` to declutter the header.

PR #390, opened 2015-12-21
Allow click events to only be sent only once

benbarnett:
Example: we want to track the number of times a user hovers
over the new navigation bar, but only need to know if they interact
with it once, rather than on every single hover.


benbarnett:
Adding directly in #389

PR #393, opened 2015-12-29
Adds `next-steps` component and applies to content pages

benbarnett:
This follows the current colour scheme, so does not include the new option colours. They will come as part of #391 

This can therefore be merged independently.

screen shot 2015-12-29 at 17 28 30


jiannis:
:ok_hand: maybe squash these three commits before merging

PR #394, opened 2015-12-30
Adds `book a free appointment` link

benbarnett:
Appears on `two-column` template and is compatible with both the current and new (dropdown nav) headers.

- [x] Create link
- [x] Use tags to determine whether to show link on relevant pages (https://github.com/guidance-guarantee-programme/pension_guidance/pull/406)

screen shot 2015-12-30 at 10 44 23

### DIFF
--- a/app/helpers/guides_helper.rb
+++ b/app/helpers/guides_helper.rb
@@ -1,0 +1,5 @@
+module GuidesHelper
+  def book_an_appointment_link?
+    !@guide.related_to_appointments?
+  end
+end
--- a/app/assets/stylesheets/components/_book-appointment.scss
+++ b/app/assets/stylesheets/components/_book-appointment.scss
@@ -1,0 +1,11 @@
+.book-appointment {
+  @include bold-19;
+  display: inline-block;
+  margin-bottom: 19px;
+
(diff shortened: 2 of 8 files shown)


andrewgarner:
The appointment pages also use the `two-column` layout so we end up linking to the current page:

book a free appointment - pension wise 20160104

benbarnett:
@andrewgarner addressed in https://github.com/guidance-guarantee-programme/pension_guidance/commit/560b75f88dfafb73a098e906597796d782b82262.

Seems cleaner than creating a whole new layout?

app/helpers/guides_helper.rb
```diff
@@ -0,0 +1,9 @@
+module GuidesHelper
+  def book_an_appointment_link?
```
benbarnett: @andrewgarner be good to get your thoughts on this... not sure what pattern we follow for helpers

andrewgarner: In Rails, all helpers are included in all views by default so I think we'll end up with a race condition. However we can configure Rails to only autoload helpers for the specific controller:

```rb
# config/application.rb
config.action_controller.include_all_helpers = false
```

PR #399, opened 2016-01-05
Inset calculators

benbarnett:
Apply visually inset style for calculators.

screen shot 2016-01-05 at 13 26 24


jiannis:
:ok_hand: maybe squash these three commits before merging

PR #400, opened 2016-01-05
Adds breadcrumb to pension option pages

benbarnett:


app/controllers/guides_controller.rb
```diff
@@ -71,5 +80,6 @@ def journey_navigation
 
   def set_breadcrumbs
     breadcrumb Breadcrumb.book_an_appointment if params[:id] == 'book'
+    breadcrumb Breadcrumb.pension_options if PENSION_OPTION_IDS.include? params[:id]
```
andrewgarner: We've now got two conditionals based on `params[:id]`. We could do this more efficiently:

```rb
case params[:id]
  when 'book'
    Breadcrumb.book_an_appointment
  when *PENSION_OPTION_IDS
    Breadcrumb.pension_options
end
```

benbarnett: Updated

PR #406, opened 2016-01-07
Tags in content frontmatter

benbarnett:
Avoids continually having to hardcode page id's to manipulate views

- [x] Add to guide model
- [x] Use in place of `PENSION_OPTION_IDS` in controller
- [x] Write more tests
- [x] Work out what to do about ordering (was previously specified by the [hardcoded array](https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/app/controllers/guides_controller.rb#L6-L20))

### DIFF
--- a/app/controllers/guides_controller.rb
+++ b/app/controllers/guides_controller.rb
@@ -1,6 +1,7 @@
 class GuidesController < ApplicationController
   layout 'guides'
   helper_method :journey_navigation
+  before_action :find_guide
   before_action :set_breadcrumbs
 
   NON_JOURNEY_RELATED_GUIDE_IDS = %w(
@@ -19,17 +20,7 @@ class GuidesController < ApplicationController
     scams
   )
 
-  PENSION_OPTION_IDS = %w(
-    adjustable-income
-    guaranteed-income
-    leave-pot-untouched
-    mix-options
-    take-cash-in-chunks
-    take-whole-pot
-  )
-
   def show
-    @guide = decorate(find(params[:id]))
     @related_guides = decorate(related_guides)
 
     expires_in Rails.application.config.cache_max_age, public: true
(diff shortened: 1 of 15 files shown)


benbarnett:
@andrewgarner all inline with master, wdyt?

app/controllers/guides_controller.rb
```diff
@@ -60,7 +38,7 @@ def related_guides
                        non_journey_related_guides
                      end
 
-    if params[:id] == 'book'
+    if tag? 'booking_form'
```
andrewgarner: Can we wrap method calls with parenthesis?

```ruby
if tag?('booking_form')
```

app/decorators/guide_decorator.rb
```diff
@@ -17,6 +17,10 @@ def content
     fail 'GuideDecorator subclasses must implement content'
   end
 
+  def tags
```
andrewgarner: We should only consider exposing the object’s tags if we need to use them in our views.

app/controllers/guides_controller.rb
benbarnett: Yep. Wonder if this is a rubocop setting too? Will take a look.

features/step_definitions/guide_steps.rb
```diff
@@ -54,27 +54,27 @@
 Then(/^I can navigate to content related to the journey$/) do
   expect(@guide.link_promo_items.map(&:text))
     .to eq(['Book a free appointment',
```
andrewgarner: The ordering of the guides is intentional.

benbarnett: Haha, ok. Will investigate.

app/repositories/guide_repository.rb
```diff
@@ -22,6 +22,12 @@ def find_all(*ids)
     end
   end
 
+  def find_tagged(tag)
```
andrewgarner: Is this needed at this time? It's not being used by a controller.

benbarnett: Nope, fair point

andrewgarner:
:+1:

PR #414, opened 2016-01-15
Remove directory from the home page

andrewgarner:
This is no longer needed since there is now a footer. The footer also adopts a different taxonomy to the directory to cause added confusion for the user!

cc/ @benbarnett


benbarnett:
:+1:

PR #415, opened 2016-01-15
Category pages

andrewgarner:


andrewgarner:
This is ready for review.

It correctly handles the taxonomy with orphaned guides populating the more category and the nested option guides listed under the title of their parent. If we update the "What you can do with your pot" title/label it be reflected here. Maybe we should add that to #418 ? 

@jiannis  - do you mind looking at the Ruby / general approach?
@benbarnett - can you have a look at styling of a link that is to an guide that is tagged with 'option'.

screen shot 2016-01-18 at 18 55 19

PR #424, opened 2016-01-22
Remove global search feature

benbarnett:
Research shows that users were confused as to whether the search was for Pension wise, GOV.UK, or both. The result was that users were being 'lost' to another website without necessarily realising.

With the addition of the dropdown navigation, we believe this is a more obvious way to navigate around the pension wise website.


andrewgarner:
:+1:

benbarnett:
@andrewgarner this is good to merge now

PR #433, opened 2016-01-28
Simplify extraction of guide headers

andrewgarner:
Always extract headers from HTML content. This ensures we can also find headers in HTML derived from Govspeak, even if there is embedded HTML in the markdown content.


benbarnett:
Seems sensible 🤘

jiannis:
Nice one!

PR #434, opened 2016-01-28
Render option circles in view not content

andrewgarner:
Where possible we should avoid presentation markup in our content files.


benbarnett:
Very yes :+1:

PR #435, opened 2016-01-28
Improve rendering of calculators

andrewgarner:
@benbarnett Am I okay making the circle a block not an inline-block so that it clears the back to description link?


benbarnett:
@andrewgarner the category page depends on them being inline-block, but we have a layout for that so can just specify it there. It's fine everywhere else, so I'll push a fix for this now.

andrewgarner:
All good to merge if you approve @benbarnett

benbarnett:
Neat :+1:

PR #436, opened 2016-01-28
Introduce calculators namespace

andrewgarner:
Provides a foundation for further calculators.


benbarnett:
:+1:

app/controllers/calculators/take_whole_pot_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Calculators
+  class TakeWholePotController < CalculatorsController
+    def show
```
jiannis: Is it OK to override the [`show`](https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/app/controllers/guides_controller.rb#L7) action defined in `GuidesController` that sets expiry cache headers?

app/controllers/calculators_controller.rb
```diff
@@ -0,0 +1,2 @@
+class CalculatorsController < GuidesController
```
jiannis: What does the `CalculatorsController` inherit from `GuidesController` ?

andrewgarner: Context of the Guide that the calculator is associated with. This allows us to display the title, breadcrumbs etc and keeps things consistent should the markdown change.

app/controllers/calculators/take_whole_pot_controller.rb
andrewgarner: That occurred to me earlier.

We could set cache headers here too, however each result is likely to be a snowflake given different pot sizes and incomes values being passed in the querystring.

app/controllers/calculators_controller.rb
jiannis: I see. Maybe at some point we could do: 

```ruby
class CalculatorsController < BaseController
  ...
end

class GuidesController < BaseController
  ...
end

class BaseController < ApplicationController
  before_action :find_page
  before_action :set_breadcrumbs

  private

  def find_page
    @page = ...
  end

  def set_breadcrumbs
    ...
  end
end

class Page
  attribute_reader :title, :description, :url, :slug
end
```

app/controllers/calculators/take_whole_pot_controller.rb
jiannis: True. Probably the traffic on these calculators won't be extreme either.

PR #437, opened 2016-01-28
Use consistent naming for calculator classes

andrewgarner:


app/assets/javascripts/take-whole-pot-calculator.js
```diff
@@ -3,7 +3,7 @@
 
   var takeWholePotCalculator = {
     init: function() {
-      var $calculator = $('.js-take-whole-pot-calculator form');
+      var $calculator = $('.js-calculator--take-whole-pot form');
```
benbarnett: This js actually [never runs](https://github.com/guidance-guarantee-programme/pension_guidance/blob/calculators/css-naming/app/assets/javascripts/application.js) in the current version, so might be simpler just to remove as we're refactoring anyway?

benbarnett:
:+1:

app/assets/javascripts/take-whole-pot-calculator.js
andrewgarner: Maybe worth doing that in another PR? I didn't want to step on the work to inline the calculators.

PR #440, opened 2016-01-28
Inline Calculators

benbarnett:
Introduces a generic javascript library to call for the calculator resulted via AJAX.


andrewgarner:
Decent.

andrewgarner:
Actually, scratch that. The behaviour if you submit an empty form is a but funky. Just recording a video to show!

benbarnett:
Good point. Hm!

andrewgarner:
It also suppresses the error states :see_no_evil:

benbarnett:
@andrewgarner have rejigged the partials to include the form with the xhr return too, so now we get error messages too.

app/views/calculators/_form_input_submit.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="form-group">
-  <input type="submit" class="button" value="Calculate" id="js-calculate">
+  <input type="submit" class="button calculator__submit js-calculate-submit" value="Calculate" id="js-calculate">
```
andrewgarner: Do we still need the id of `js-calculate`?

app/controllers/calculators/take_whole_pot_controller.rb
```diff
@@ -5,6 +5,10 @@ def show
       @income = params[:income]
 
       @calculator = TakeWholePotCalculatorForm.new(pot: @pot, income: @income)
+
+      render partial: 'calculators/take_whole_pot/calculator',
+             locals: { calculator: @calculator, pot: @pot, income: @income },
+             status: (@calculator.invalid? ? 400 : 200) if request.xhr?
```
andrewgarner: [Rails understands both numeric status codes and corresponding symbols](http://guides.rubyonrails.org/layouts_and_rendering.html).

I think using symbols improves the readability if you're not hot on your HTTP status codes.

```rb
status: (@calculator.valid? ? :ok : :bad_request)
```

app/views/calculators/take_whole_pot/_calculator.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= render 'form', calculator: @calculator, pot: @pot, income: @income %>
```
andrewgarner: You should use the locale variables passed to the partial not the instance variables defined by the controller. This allows the partial to be used in any context without it needing to know how it is being used.

```rb
<%= render 'form', calculator: calculator, pot: pot, income: income %>
<%= render 'result', calculator: calculator %>
```

andrewgarner:
:+1:

PR #441, opened 2016-01-28
Changes to Tax.md

NikkiAlbon:
Changed pension recycling information in line with new fact checks from HMRC. Took out factually incorrect information. Shortened the information to just the basics because the detail of the information is too confusing to understand on a web page. Guidance is to see a financial adviser if you're thinking of paying in money to your pension at the same time as taking it out.

### DIFF
--- a/content/tax.md
+++ b/content/tax.md
@@ -129,14 +129,8 @@ You’ll have to pay tax on any amount in your pension pot above your [lifetime 
 
 ##Pension recycling
 
-If you take a lump sum from your pension and you use that lump sum to pay money into another pension pot in the same tax year, it’s known as ‘pension recycling’.
-
-If you do this you may have to pay 55% tax on your pension contribution.
-
-You may have to pay the tax if any money you put into another pension pot:
-
-- exceeds £7,500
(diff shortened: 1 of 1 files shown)


content/tax.md
```diff
@@ -129,14 +129,8 @@ You’ll have to pay tax on any amount in your pension pot above your [lifetime
...
-If you take a lump sum from your pension and you use that lump sum to pay money into another pension pot in the same tax year, it’s known as ‘pension recycling’.
-
-If you do this you may have to pay 55% tax on your pension contribution.
-
-You may have to pay the tax if any money you put into another pension pot:
-
-- exceeds £7,500
-- exceeds 1% of the lifetime allowance before 6 April 2015
-- increases your usual pension contributions by 30% or more
+If you take a tax-free lump sum from your pot and you pay into the same or another pot, it’s known as ‘pension recycling’.
```
andrewgarner: My question was going to be "Why would you take money from your pot to then pay it back into the same pot?"

I thought that because we've lost the context of "in the same tax year" that was there before.

PR #442, opened 2016-01-29
Improve calculator error handling

andrewgarner:
If there are multiple errors for an attribute we should show them, however, we can display error messages that give more context about the specific error instead of just using the same error messages for all validation errors.


benbarnett:
Beaut :+1:

PR #446, opened 2016-02-01
Simplify print styles for calculators

benbarnett:
screen shot 2016-02-01 at 10 19 02

### DIFF
--- a/app/assets/javascripts/calculators.js
+++ b/app/assets/javascripts/calculators.js
@@ -16,6 +16,7 @@
     },
 
     updateResults: function(data) {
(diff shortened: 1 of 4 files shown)


andrewgarner:
Can the £ signs be bold for the first two figures so they are consistent?

What happens if the user hasn't completed the calculator? Do we still print it?

benbarnett:
@andrewgarner updated. The input numbers are now all standard weight, and the results are the only numbers in bold. Seems clearer.

The calculator only shows on print media after the first successful result.

andrewgarner:
:+1:

PR #452, opened 2016-02-02
Phone vs face to face appointments

jiannis:
Different pages for booking phone or face to face appointments for Tim's experiment.

Ask Tim if this is ready to go live

:two_men_holding_hands:

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,6 +4,8 @@ Rails.application.routes.draw do
   constraints format: 'html' do
     resources :categories, only: 'show', path: 'browse'
 
+    get '/book', to: redirect('/appointments', status: 302)
+
     GuideRepository.new.all.each do |guide|
       get guide.slug,
           controller: 'guides',
--- a/content/appointments.md
+++ b/content/appointments.md
@@ -10,7 +10,8 @@ An appointment is a 45 minute conversation between you and a Pension Wise guidan
 
 Appointments are over the **phone** or **face to face**.
 
-[Book now](/book){: .button .button-get-started}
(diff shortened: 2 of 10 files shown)


content/appointments.md
```diff
@@ -10,7 +10,8 @@ An appointment is a 45 minute conversation between you and a Pension Wise guidan
 
 Appointments are over the **phone** or **face to face**.
 
-[Book now](/book){: .button .button-get-started}
+[Book a telephone appointment](/book-phone){: .button #phone-button style="font-size: 0.9em;"}
```
andrewgarner: Shouldn't it say "Book a **phone** appointment"?

config/routes.rb
```diff
@@ -4,6 +4,8 @@
   constraints format: 'html' do
     resources :categories, only: 'show', path: 'browse'
 
+    get '/book', to: redirect('/book-phone')
```
andrewgarner: This will create a 301 "Moved Permanently" redirect which we don't want. We should do this upstream where we can set a 302 redirect.

jiannis: :ok_hand: good point

jiannis: I've changed the redirect status to 302 since [it's possible] (https://github.com/rails/rails/blob/69c7c06923249e7c95491db33933bbafa19707d3/actionpack/lib/action_dispatch/routing/redirection.rb#L179) in f6649f2 instead of doing it upstream so as not to change the behaviour of [this feature test](https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/features/guides.feature#L15).

jiannis:
@andrewgarner squash the commits to be easier to revert after the “experiment” is over?

PR #453, opened 2016-02-03
Add GTM custom event `CalculatorEstimate`

benbarnett:
The GTM tag will use a trigger that listens for this event.

### DIFF
--- a/app/views/calculators/leave_pot_untouched/_form.html.erb
+++ b/app/views/calculators/leave_pot_untouched/_form.html.erb
@@ -27,6 +27,7 @@
   <script>
     window.dataLayer = window.dataLayer || [];
     dataLayer.push({
+      'event': 'CalculatorEstimate',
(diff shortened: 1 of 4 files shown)


app/views/calculators/leave_pot_untouched/_form.html.erb
```diff
@@ -27,6 +27,7 @@
   <script>
     window.dataLayer = window.dataLayer || [];
     dataLayer.push({
+      'event': 'CalculatorResult',
```
andrewgarner: Do you think we should call it `CalculatorResult`?

I'm inclined to rename the partials as we've adopted the path `/foo-calculator/estimate`.

benbarnett: Yep, might be cleaner.

andrewgarner:
:+1:

PR #459, opened 2016-02-04
Round calculator estimates to the nearest whole number

### LINKED ISSUE #458: Always show whole numbers on calculators
Requested by UX, seems to simplify the view.

benbarnett:
Closes #458.


app/helpers/application_helper.rb
```diff
@@ -2,7 +2,7 @@ module ApplicationHelper
   include NavigationHelper
 
   def format_currency(amount, **options)
-    number_to_currency(amount, options)&.sub(/\.00$/, '')
+    number_to_currency(amount, options.reverse_update(precision: 0))&.sub(/\.00$/, '')
```
andrewgarner: I think you can set precision in the `en.yml` file.

`&.sub(/\.00$/, '')` is also now redundant.

benbarnett: Agree. Might we want this behaviour at some point though?

andrewgarner: We can re-introduce it if we need it.

benbarnett: Cool. Have cleaned it up now.

andrewgarner:
:+1:

andrewgarner:
Actually ignore my thumb - one more comment

app/helpers/application_helper.rb
```diff
@@ -2,7 +2,7 @@ module ApplicationHelper
   include NavigationHelper
 
   def format_currency(amount, **options)
-    number_to_currency(amount, options)&.sub(/\.00$/, '')
+    number_to_currency(amount, options)
```
andrewgarner: We might as well remove `format_currency` completely as all it now does is proxy to `number_to_currency`. We could just use `number_to_currency` in our views and keep things closer to the Rails norm.

andrewgarner:
:+1:

benbarnett:
Ta. Much better.

PR #460, opened 2016-02-04
Extract reusable income tax calculator

andrewgarner:
The Take Whole Pot calculator calculates income tax based upon income and the total value of a pension pot. The total value of a pension pot is really just a lump sum in the context of how it is used.

This class doesn’t need to know where the lump sum came from, it should just calculate income tax in any given context.

This change makes the class reusable.


benbarnett:
:+1

PR #469, opened 2016-02-16
Add higher rate tax warning to cash in chunks page

benbarnett:


andrewgarner:
This is live, however, we have two very similar warnings that are treated differently. One about benefits and one about higher rate tax.

screen shot 2016-02-22 at 08 37 15

What do you think @benbarnett @guidance-guarantee-programme/content-designers ? Maybe we could combine them into one warning – though they cover different ends of the spectrum?

> Taking a large sum of cash from your pot can mean you pay a higher amount of tax and could also affect your entitlement to any benefits.

andrewgarner:
Addressed with #470

PR #472, opened 2016-02-24
Adjustable income calculator

benbarnett:
screen shot 2016-04-11 at 10 50 41

### DIFF
--- a/app/assets/javascripts/calculators.js
+++ b/app/assets/javascripts/calculators.js
@@ -3,6 +3,7 @@
 
   var calculators = {
     scrollSpeed: 700,
+    request: null,
 
     init: function() {
       this._initDOM();
@@ -12,12 +13,20 @@
     submitForm: function() {
       var $form = this.$calculator.find('form');
       this._toggleLoading(true);
-      return $.get($form.attr('action'), $form.serialize());
+
+      if (this.request && typeof this.request.abort === 'function') {
+        this.request.abort();
+      }
+
+      this.request = $.get($form.attr('action'), $form.serialize());
+
+      return this.request;
     },
 
     updateEstimate: function(data) {
       this.$calculator.removeClass('hide-from-print');
       this._refresh(data);
+      this._sliders();
       this._scrollTo(this.$scrollTarget).then($.proxy(function() {
         this._toggleLoading(false);
       }, this));
@@ -30,8 +39,15 @@
       }, this));
     },
 
-    _refresh: function(html) {
-      this.$calculator.empty().html(html);
+    _refresh: function(html, $partial) {
+      var $target = this.$calculator;
+
+      if ($partial) {
+        $target = $partial;
+      }
+
+      $target.empty().html(html).find('.calculator__highlight').removeClass('highlight').addClass('highlight');
+
       this._cacheElements();
     },
 
@@ -45,6 +61,11 @@
 
       this.$submitButton = this.$calculator.find('.js-calculate-submit');
       this.$estimate = this.$calculator.find('.js-calculator-estimate');
+
+      if (this.$loadingStatus) {
+        this.$loadingStatus.remove();
+      }
+
       this.$loadingStatus = $('<span class="calculator__loading-status">Please wait...</span>').
                            insertAfter(this.$submitButton);
       this.$scrollTarget = this.$submitButton;
@@ -63,8 +84,39 @@
       }, this));
     },
 
+    _updateEstimateOnly: function() {
+      this.submitForm().then($.proxy(function(data) {
+        var $empty = $('<div/>');
+        var $estimate = $empty.append(data).find('#js-estimate');
+        this._refresh($estimate, $('#js-estimate'));
+
+        this._scrollTo(this.$scrollTarget).then($.proxy(function() {
+          this._toggleLoading(false);
+        }, this));
+
+      }, this));
+    },
+
     _toggleLoading: function(isLoading) {
       this.$loadingStatus[isLoading ? 'addClass' : 'removeClass']('calculator__loading-status--loading');
+    },
+
+    _sliders: function() {
+      if ($('#slider').length) {
+        var slider = new PWPG.Slider($('#slider'));
+        var $target = $($('#slider').attr('data-target'));
+        var buffer;
+
+        slider.$textInput.on('change', $.proxy(function() {
+          clearTimeout(buffer);
+
+          buffer = setTimeout($.proxy(function() {
+            $target.val(slider.$textInput.val());
+            this._updateEstimateOnly();
+          }, this), 200);
+
+        }, this));
+      }
     },
 
     _scrollTo: function($el) {
--- a/app/assets/javascripts/calculators.js
+++ b/app/assets/javascripts/calculators.js
@@ -121,9 +121,18 @@
 
     _scrollTo: function($el) {
       var offset = $el.offset() || {};
-      return $('html, body').animate({
-        scrollTop: offset.top || 0
-      }, this.scrollSpeed).promise();
+      var target = offset.top || 0;
+      var $page = $('html, body');
+
+      $page.on('scroll mousedown wheel DOMMouseScroll mousewheel touchmove', function() {
+        $page.stop();
+      });
+
+      return $page.animate({
+        scrollTop: target
+      }, this.scrollSpeed).promise().then(function() {
+        $page.off('scroll mousedown wheel DOMMouseScroll mousewheel touchmove');
+      });
     }
   };
 
--- a/app/controllers/calculators/adjustable_income_controller.rb
+++ b/app/controllers/calculators/adjustable_income_controller.rb
@@ -1,0 +1,27 @@
+module Calculators
+  ARBITRARY_MULTIPLIER = 5
+
+  class AdjustableIncomeController < CalculatorsController
+    def show
+      @form = AdjustableIncomeForm.new(form_params)
+
+      if request.xhr?
+        status = @form.invalid? ? :bad_request : :ok
+
+        render partial: 'calculators/adjustable_income/calculator',
+               locals: { form: @form },
+               status: status
+      end
+    end
+
+    private
+
+    def id
+      'adjustable-income'
+    end
+
+    def form_params
+      params.permit(:pot, :age, :desired_monthly_income)
+    end
+  end
+end
--- a/app/forms/calculators/adjustable_income_form.rb
+++ b/app/forms/calculators/adjustable_income_form.rb
@@ -1,0 +1,43 @@
+require 'active_model'
+
+module Calculators
+  class AdjustableIncomeForm
+    include ActiveModel::Model
+
+    attr_reader :pot, :age, :desired_monthly_income
+
+    validates :pot, presence: true, numericality: { allow_blank: true,
+                                                    greater_than: 0,
+                                                    less_than: 5_000_000 }
+    validates :age, numericality: { allow_blank: true, greater_than_or_equal_to: 55 }
+    validates :desired_monthly_income,
+              numericality: { allow_blank: true, greater_than: 0,
+                              less_than: -> (form) { form.errors.include?(:pot) ? Float::INFINITY : form.pot.to_f } }
+
+    def pot=(pot)
+      @pot = String(pot).delete(',').squish
+    end
+
+    def age=(age)
+      @age = String(age).delete(',').squish
+    end
+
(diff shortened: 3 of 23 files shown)


app/views/calculators/adjustable_income/_calculator.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= render 'form', form: form %>
+<%= render 'estimate', estimate: form.estimate, pot: form.pot, monthly_drawdown_amount: form.estimate[:monthly_drawdown_amount] if form.valid? %>
```
benlovell: The postfix modifier is a surprise here since the line is reasonably long. I think this could be clearer at a glance if we move the `render` call inside the conditional instead.

features/step_definitions/calculator_steps.rb
```diff
@@ -52,12 +71,33 @@
...
+Then(/^I should see an income that will last until my life expectancy$/) do
+  expect(@page.calculator.desired_income_lasts_until).to have_content('84')
+end
+
+Then(/^it shows me what age my desired income will last until$/) do
+  expect(@page.calculator.desired_income_lasts_until).to have_content('70')
+end
+
+And(/^it shows me a suggested monthly income that will last until an average life expectancy for someone of my age$/) do
+  content = 'An average life expectancy for someone born in the same year as you is 84 years. '\
```
benlovell: Matching on this is pretty fragile as-is. The important stuff here is the years and monthly income, so perhaps we could just match on those with a permissive regex for the rest?

features/step_definitions/calculator_steps.rb
```diff
@@ -52,12 +71,33 @@
...
+            'You could take a monthly income of £325 which would last until then.'
+
+  expect(@page.calculator.income_for_life_expectancy).to have_content(content)
+end
+
 Then(/^it explains the values are estimates based on growth at 3% per year$/) do
   content = 'This is an estimate based on your whole pot growing at a rate of about 3% per year — this may vary.'
```
benlovell: Same here with regards matching the whole content. The important bit is the growth interest rate.

features/step_definitions/calculator_steps.rb
```diff
@@ -52,12 +71,33 @@
...
 
   expect(@page.calculator.notes).to have_content(content)
 end
 
+Then(/^it explains the values are estimates based on the amount remaining growing at 3% per year$/) do
+  content = 'This is an estimate based the amount in your pot growing at a rate of about 3% per year — this may vary.'
```
benlovell: Same here with regards matching the whole content. The important bit is the growth interest rate.

lib/adjustable_income_calculator.rb
```diff
@@ -0,0 +1,99 @@
+class AdjustableIncomeCalculator
+  # https://www.ons.gov.uk/peoplepopulationandcommunity/birthsdeathsandmarriages/lifeexpectancies/bulletins/lifeexpectancyatbirthandatage65bylocalareasinenglandandwales/2015-11-04#national-life-expectancy-at-age-65
```
benlovell: A commit message on the commit that introduced this code would maybe be more suitable for this particular reference?

lib/adjustable_income_calculator.rb
```diff
@@ -0,0 +1,99 @@
...
+      desired_income_with_pot_growth_lasts_until: desired_income_with_pot_growth_lasts_until.to_i,
+      monthly_income_until_life_expectancy: monthly_income_until_life_expectancy.to_i,
+      life_expectancy: LIFE_EXPECTANCY
+    }
+  end
+
+  private
+
+  def normalised_drawdown_amount
+    (desired_income > 0) ? desired_income : income_until_life_expectancy
```
benlovell: Since Ruby 2.3 numerics now have a `positive?` predicate. That would remove the operator / operand here and improve readability a teeny bit.

lib/adjustable_income_calculator.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def years_pot_will_last_for(yearly_drawdown, pot)
+    (pot / yearly_drawdown).floor
+  end
+
+  def desired_income_with_pot_growth_lasts_until(yearly_drawdown = normalised_drawdown_amount)
+    pot_remaining = taxable_portion
+    years_lasted = 0
+
+    while pot_remaining > 0
```
benlovell: Same here and elsewhere in this method with regards replacing the operator / operand with the `positive?` predicate.

app/controllers/calculators/adjustable_income_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Calculators
+  class AdjustableIncomeController < CalculatorsController
+    def show
+      @form = AdjustableIncomeForm.new(form_params)
+
+      render partial: 'calculators/adjustable_income/calculator',
+             locals: { form: @form },
+             status: (@form.invalid? ? :bad_request : :ok) if request.xhr?
```
benlovell: Promoting that postfix modifier to a regular conditional would clarify this a little. I think it's hidden away over there.

Also, extracting that ternary expression out to a separate line / assignment makes this easier to scan / parse.

app/forms/calculators/adjustable_income_form.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def desired_income
+      @desired_monthly_income.to_f * 12
+    end
+
+    def estimate
+      AdjustableIncomeCalculator.new(pot: pot.to_f,
+                                     age: age.to_f,
+                                     desired_income: desired_income.to_f).estimate if valid?
```
benlovell: Same here with the postfix modifier.

benlovell:
Looks pretty solid :+1: Just minor stylistic stuff mostly. There is a general theme around using postfix modifiers that could be contentious, perhaps we could have that discussion and update the rubocop rules if we decided one way or the other?

app/views/calculators/adjustable_income/_estimate.html.erb
```diff
@@ -0,0 +1,38 @@
...
+
+    <strong class="calculator__estimate__number t-calculator-tax-free-lump-sum">
+      <%= number_to_currency(estimate[:tax_free_lump_sum]) %> tax free
+    </strong>
+
+    and from the <span class="t-calculator-taxable-portion"><%= number_to_currency(estimate[:taxable_portion]) %></span> left over:
+
+    <strong class="calculator__estimate__number">
+      a monthly income of <span class="calculator__highlight"><%= number_to_currency(monthly_drawdown_amount) %></span>
+      would last<br>until you're <span class="t-calculator-age-for-desired-income calculator__highlight"><%= estimate[:desired_income_with_pot_growth_lasts_until] %></span>
```
andrewgarner: `you’re`

app/views/calculators/adjustable_income/_estimate.html.erb
```diff
@@ -0,0 +1,38 @@
...
+<p>
+  <label for="slider">or try with a different monthly income:</label>
+  <%= render 'components/slider', {
+      id: 'slider',
+      target: '#desired_monthly_income',
+      option: 'adjustable-income',
+      text_before: '£',
+      initial_value: monthly_drawdown_amount,
+      min: 50,
+      max: monthly_drawdown_amount.to_f * 5,
```
andrewgarner: 5 is a magic number

app/views/calculators/adjustable_income/_form.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+<% if form %>
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    dataLayer.push({
+      'event': 'CalculatorEstimate',
+      'valid': <%= form.valid? %>
+      <% if form.valid? %>
+      , 'pot': <%= form.pot.to_i %>
+      , 'age': <%= form.age.to_i %>
```
andrewgarner: Do you want to surface the `income` value too?

features/step_definitions/calculator_steps.rb
andrewgarner: :+1: — this has caught us out before when the @guidance-guarantee-programme/content-designers have tweaked content and broken the build.

lib/adjustable_income_calculator.rb
andrewgarner: :+1:

lib/adjustable_income_calculator.rb
andrewgarner: :+1:

lib/adjustable_income_calculator.rb
```diff
@@ -0,0 +1,99 @@
...
+    amount_with_growth(taxable_portion, years_pot_will_last_without_growth, normalised_drawdown_amount)
+  end
+
+  def income_until_life_expectancy
+    yearly_withdrawal_until_death = desired_income
+
+    yearly_withdrawal_until_death = pot if yearly_withdrawal_until_death <= 0
+
+    while desired_income_with_pot_growth_lasts_until(yearly_withdrawal_until_death) < LIFE_EXPECTANCY
+      yearly_withdrawal_until_death -= 10
```
andrewgarner: `10` is a magic number

app/views/calculators/adjustable_income/_estimate.html.erb
dwhenry: Should we add logic so that the min value can not result in payments until your 150. :)

app/assets/javascripts/calculators.js
```diff
@@ -56,15 +77,55 @@
...
     _scrollTo: function($el) {
       var offset = $el.offset() || {};
+      var target = offset.top || 0;
+      var $page = $('html, body');
+
+      $page.on('scroll mousedown wheel DOMMouseScroll mousewheel touchmove', function() {
+        $page.stop();
+      });
+
       return $('html, body').animate({
```
dwhenry: use $page variable

benbarnett: good spot

lib/adjustable_income_calculator.rb
```diff
@@ -0,0 +1,99 @@
+class AdjustableIncomeCalculator
+  # https://www.ons.gov.uk/peoplepopulationandcommunity/birthsdeathsandmarriages/lifeexpectancies/bulletins/lifeexpectancyatbirthandatage65bylocalareasinenglandandwales/2015-11-04#national-life-expectancy-at-age-65
+  LIFE_EXPECTANCY = 84
+  TAX_FREE_POT_PORTION = 0.25
+  GROWTH_INTEREST_RATE = 0.03
+
+  def initialize(pot:, age:, desired_income:)
```
benlovell: Depending on the size of these and that they're unbounded, it's quite trivial to pretty much DoS the server.

We can take a closer look at this tomorrow. It might be a simple case of just validating that the pot size in particular does not exceed a super-huge amount.

I do wonder whether there may be a domain-specific reason why this wouldn't generally exceed say £10,000,000 or indeed any other arbitrarily large amount. Perhaps in this case the end-user is likely to have specialised accountancy  and wouldn't likely use this calculator?

lib/adjustable_income_calculator.rb
benlovell: I also found tweaking this number higher could dramatically reduce the wall time taken on this calculation, yet didn't overly affect the accuracy.

app/views/calculators/adjustable_income/_form.html.erb
benbarnett: We'll need to send this through as `pot_payout` (`income` is only related to the user's current earnings).

So essentially `, 'pot_payout': monthly_drawdown_amount.to_i`

benbarnett: Having said that, `monthly_drawdown_amount` isn't available in this context. We can send a separate event to the datalayer on the `_estimate.html.erb` partial though:

```erb
<script>
  window.dataLayer = window.dataLayer || [];
  dataLayer.push({
    'event': 'CalculatorEstimate',
    'pot_payout': monthly_drawdown_amount.to_i
  });
</script>
```

lib/adjustable_income_calculator.rb
dwhenry: not sure if this is helpful: http://www.financeformulas.net/Loan_Payment_Formula.html

app/views/calculators/adjustable_income/_form.html.erb
benbarnett: Have added this to `_estimate.html.erb`

PR #484, opened 2016-03-08
Use `sassc-rails` for sass compilation

benbarnett:
This uses the libsass bindings which reduces compilation time in the asset pipeline quite significantly. Also enables source maps in the development environment.


andrewgarner:
:+1:

andrewgarner:
Should go hand in hand with https://github.com/guidance-guarantee-programme/pension_guidance/pull/482 for the double whammy.

PR #490, opened 2016-03-11
Redrafted text and moved calculator

CassBonner:
Redrafted text and moved calculator after testing showed users were confused by mixed messaging of 'leave untouched' and 'pay in'.

Would also like to add a 3rd bullet to in-calculator text:

- You must leave your whole pot – if you take a 25% tax-free lump sum, you must take the rest of your money within 6 months, using one of the [other pension options](/pension-pot-options).

### DIFF
--- a/content/leave_pot_untouched.md
+++ b/content/leave_pot_untouched.md
@@ -13,23 +13,26 @@ You don’t have to start taking money from your pension pot when you reach your
 
 - The money in your pot could grow.
 - You could have more money to last a shorter length of time.
-- You (and your employer) can also still pay into your pot but there may be restrictions.
-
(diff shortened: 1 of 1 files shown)


benbarnett:
@CassBonner it looks like that bullet point might already be there. Does this look right to you?  https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/app/views/calculators/leave_pot_untouched/_estimate.html.erb#L23-L34

CassBonner:
That's weird - it's not on the live site. No matter...that looks good 😀

NikkiAlbon:
Hi Cass,

I had added that as part of another card.

Nikki

PR #491, opened 2016-03-11
Slider component

benbarnett:
Just added to the styleguide for now.

screen shot 2016-03-14 at 14 46 24


andrewgarner:
How about adding [ESLint](https://github.com/babel/babel-eslint)?

benbarnett:
Not a bad idea. What do you think about some acceptance tests for rendering the [slider partial](https://github.com/guidance-guarantee-programme/pension_guidance/blob/slider/app/views/components/_slider.html.erb) itself? Overkill?

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,88 @@
...
+    background: $track-fill no-repeat $track-bg;
+    background-size: 0 100%;
+  }
+}
+
+@mixin thumb() {
+  width: $thumb-d;
+  height: $thumb-d;
+  border-radius: $thumb-d;
+  border: 3px solid $color-option-leave-pot-untouched;
```
andrewgarner: We shouldn't be using an option specific colour here. If the slider changes when used in the context of an option we should modify the colour in context.

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,88 @@
+$track-w: 16em;
+$track-h: .625em;
+$track-bg: $grey-4;
+$track-fill: linear-gradient(90deg, $grey-2 25%, $grey-2 100%);
+
+$thumb-d: 1.5em;
```
andrewgarner: What does `-d` mean? Can we use more expressive variable names like `$track-width`, `$track-height` etc?

app/assets/stylesheets/components/_slider.scss
benbarnett: Yeah – need to run the whole thing by design tbh.

app/assets/stylesheets/components/_slider.scss
benbarnett: It means diameter. Agree, but will wait until design have taken a look before finalising the CSS.

andrewgarner: No problem.

On Mon, 14 Mar 2016 at 10:03 Ben Barnett <notifications@github.com> wrote:

> In app/assets/stylesheets/components/_slider.scss
> <https://github.com/guidance-guarantee-programme/pension_guidance/pull/491#discussion_r55977063>
> :
>
> > @@ -0,0 +1,88 @@
> > +$track-w: 16em;
> > +$track-h: .625em;
> > +$track-bg: $grey-4;
> > +$track-fill: linear-gradient(90deg, $grey-2 25%, $grey-2 100%);
> > +
> > +$thumb-d: 1.5em;
>
> It means diameter. Agree, but will wait until design have taken a look
> before finalising the CSS.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guidance-guarantee-programme/pension_guidance/pull/491/files#r55977063>
> .
>
-- 
Andrew Garner
*Pension Wise*

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,199 @@
...
+$track-fill: linear-gradient(90deg, $grey-3 25%, $grey-3 100%);
+
+$thumb-diameter: 1.5em;
+
+$option-colours: (
+  "leave-pot-untouched": $color-option-leave-pot-untouched,
+  "guaranteed-income": $color-option-guaranteed-income,
+  "adjustable-income": $color-option-adjustable-income,
+  "take-cash": $color-option-take-cash,
+  "whole-pot": $color-option-take-whole-pot,
```
andrewgarner: `"take-whole-pot"` ?

benbarnett: Agree, but this is consistent with the other CSS components so am tempted to keep as is and maybe fix in a separate PR?

andrewgarner: Ahh OK - yep, we should fix it as one.

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,199 @@
+$track-width: 16em;
+$track-height: .625em;
+$track-bg: $white;
```
andrewgarner: `$track-background`?

benbarnett: Fixed

benbarnett:
@andrewgarner have added eslint in https://github.com/guidance-guarantee-programme/pension_guidance/commit/a7abefc6d671e910a2917e04d44ece4e1276c6b9

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,199 @@
...
+  border-bottom: 2px solid $grey-3;
+  background: $track-bg;
+
+  @if $flag {
+    background: $track-fill no-repeat $track-bg;
+    background-size: 0 100%;
+  }
+}
+
+@mixin thumb($option: "leave-pot-untouched") {
```
andrewgarner: Do we want to couple the slider component to being used in the context of an option?

If we passed in the colour instead of the map key we could remove the coupling.

benbarnett: Good point, will rework

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,199 @@
+$track-width: 16em;
```
andrewgarner: Is there a reason why we're defining the variables here and not in https://github.com/guidance-guarantee-programme/pension_guidance/blob/slider/app/assets/stylesheets/lib/_variables.scss ?

benbarnett: Nope and just noticed this myself! Oops. Will move in.

app/assets/stylesheets/components/_slider.scss
```diff
@@ -0,0 +1,199 @@
...
+    @include thumb();
+  }
+
+  &:focus {
+    outline: 2px solid $focus-colour;
+  }
+}
+
+.slider--leave-pot-untouched {
+  .slider__range-input {
```
andrewgarner: Is it possible to extract `.slider__range-input` into something that can be reused so that we don't have to duplicate it for each option variation?

benbarnett: Yes, but would question readability over SASS optimisation at this stage. Ultimately this selector needs to be in the output CSS which I think is made clearer using this method. In fact all of these option variants could be rendered using a SASS `each` loop, which would DRY up the code completely. But, I feel that working with SASS code that is closer to the final CSS output is a good way of doing things... what dya reckon?

app/views/styleguide/components/_slider.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render 'components/slider', { option: 'leave-pot-untouched', initial_value: rand(0..150) } %>
```
andrewgarner: Maybe include an example for the default slider too? (i.e. with no `option` key defined)

app/assets/javascripts/components/Slider.es6
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    createSlider() {
+      this.$textInput = this.$component.find('[data-slider-text-input]');
+      this.$rangeInput = this.$textInput
+            .clone()
+            .addClass('slider__range-input')
+            .removeAttr('data-slider-text-input name')
+            .attr({
+              'type'      : 'range',
```
andrewgarner: The key value whitespace formatting is inconsistent here.

benbarnett: Fixed

app/views/styleguide/components/_slider.html.erb
benbarnett: Good idea, have added

andrewgarner:
:+1:

PR #498, opened 2016-03-22
Digital Summary document

benbarnett:
Gives customers who have had an appointment the ability to generate a digital version of the summary document they receive in the post.


app/views/appointment_summaries/new.html.erb
```diff
@@ -0,0 +1,156 @@
...
+<script type="text/html" id="consent">
+  <div class="checkbox">
+    <label for="appointment_summary_consent">
+      <input class="t-continue-working" type="checkbox" value="1" name="appointment_summary[consent]" id="appointment_summary_consent" />
+      Visitor has consented to sharing data for purpose of creating and sending summary document
+    </label>
+  </div>
+</script>
+
+<script type="text/html" id="postcode-lookup-template">
```
dwhenry: Is this used on this page?

benbarnett: Nope, good point!

Gemfile
```diff
@@ -31,6 +34,7 @@ gem 'sprockets-rails', '~> 3.0'
 gem 'uglifier', '>= 1.3.0'
 gem 'uk_postcode'
 gem 'utf8-cleaner'
+gem 'output-templates', '~> 4.1.0', github: 'guidance-guarantee-programme/output-templates', branch: 'web-variant'
```
andrewgarner: Alphabetical ordering :)

app/assets/stylesheets/components/_sticky-sidebar.scss
```diff
@@ -194,3 +194,10 @@
     }
   }
 }
+
+.sticky-sidebar--summary-document {
```
andrewgarner: `.sticky-sidebar--appointment-summary`?

app/assets/stylesheets/components/summary-document.scss
```diff
@@ -0,0 +1,75 @@
+//*-------------- mixins --------------------*/
+@mixin bullet($color, $radius, $top: 5px) {
```
andrewgarner: Should we be using or evolving `components/_bullet.scss`?

benbarnett: Yeah, I thought about this – it would involve changing the markup in `output-templates` though. So will run this through with content/design first and revisit I think.

app/controllers/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+      format.html { render html: output_document.html.html_safe }
+      format.pdf do
+        send_data output_document.pdf,
+                  filename: 'pension_wise.pdf', type: 'application/pdf',
+                  disposition: :inline
+      end
+    end
+  end
+
+  def view
```
andrewgarner: Why not `show`?

app/controllers/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  layout 'guides'
+
+  helper_method :headers
+  helper_method :appointment_type
+
+  def new
+    @appointment_summary = AppointmentSummary.new
+  end
+
+  def create
```
andrewgarner: We're not actually creating anything so I'm not sure this the correct verb to use?

We're requesting an appointment summary, either in HTML or PDF and we need to supply specific params for the request to be accepted.

Gemfile
benbarnett: Fixed

app/assets/stylesheets/components/_sticky-sidebar.scss
benbarnett: Changed

app/controllers/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+    appointment_summary = AppointmentSummary.new(appointment_summary_params)
+    output_document = OutputDocument.new(appointment_summary)
```
andrewgarner: Maybe we could rename `OutputDocument` to something that makes more sense within the domain terms of this application?

`OutputDocument` is essentially a [factory](https://practicingruby.com/articles/creational-design-patterns) – it allows us to manufacture a HTML or PDF representation of the appointment summary but without needing knowledge of how the formats are created.

app/domain/output_document/html_renderer.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def render
+      template.render(output_document)
+    end
+
+    private
+
+    def template
+      template_id = (output_document.variant == 'other') ? 'ineligible' : output_document.variant
```
andrewgarner: Varient is always `web` so this is redundant.

app/views/appointment_summaries/new.html.erb
```diff
@@ -0,0 +1,156 @@
+<%= form_for @appointment_summary, url: view_appointment_summaries_path do |f| %>
```
andrewgarner: `view_appointment_summaries_path` feels like it should just be `appointment_summary_path`?

app/models/appointment_summary.rb
```diff
@@ -0,0 +1,31 @@
+class AppointmentSummary
+  include ActiveModel::Model
+
+  attr_accessor :supplementary_benefits, :supplementary_debt, :supplementary_ill_health,
+                :supplementary_defined_benefit_pensions,
+                :appointment_type, :format_preference
```
andrewgarner: Does `format_preference` need to be part of the public interface given we're not using it?

app/models/appointment_summary.rb
```diff
@@ -0,0 +1,30 @@
+class AppointmentSummary
+  include ActiveModel::Model
+
+  attr_accessor :supplementary_benefits, :supplementary_debt, :supplementary_ill_health,
+                :supplementary_defined_benefit_pensions, :appointment_type
+
+  %i(supplementary_benefits supplementary_debt
+     supplementary_ill_health supplementary_defined_benefit_pensions).each do |predicate_method|
```
benlovell: `predicate_method` as a name doesn't make sense here. They're attribute writers, not predicates.

app/models/appointment_summary.rb
```diff
@@ -0,0 +1,30 @@
...
+    define_method("#{predicate_method}=") do |value|
+      boolean = [true, 'true', '1', 1].member?(value)
+      instance_variable_set("@#{predicate_method}", boolean)
+    end
+
+    alias_method(:"#{predicate_method}?", predicate_method)
+  end
+
+  def initialize(params = {})
```
benlovell: You can just call `super` and it passes along the same arguments.

app/models/appointment_summary.rb
benbarnett: Renamed

app/models/appointment_summary.rb
benbarnett: Changed

app/controllers/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      render :new
+    end
+  end
+
+  def download
+    @appointment_summary = AppointmentSummary.new(appointment_summary_params)
+    output_document = OutputDocument.new(@appointment_summary, 'generic')
+
+    send_data output_document.pdf,
+              layout: false,
```
benlovell: The `layout` option is redundant here.

app/models/appointment_summary.rb
```diff
@@ -0,0 +1,30 @@
+class AppointmentSummary
+  include ActiveModel::Model
+
+  attr_accessor :supplementary_benefits, :supplementary_debt, :supplementary_ill_health,
```
dwhenry: We repeat this list of fields a lot, maybe a constant to hold the list would reduce duplication?

app/controllers/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def content_document
+    @content_document ||= Nokogiri::HTML(content)
+  end
+
+  def appointment_summary_params
+    params
+      .require(:appointment_summary)
+      .permit(:supplementary_benefits, :supplementary_debt,
```
benlovell: Could we favour the style where each element, including the opening and closing paren is on a new line? It makes the diff more readable and helps if we add / remove others.

app/controllers/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  def print
+    @appointment_summary = AppointmentSummary.new(appointment_summary_params)
+    output_document = OutputDocument.new(@appointment_summary, 'generic')
+
+    @content = output_document.html.html_safe
+  end
+
+  private
+
+  def headers(level = 1)
```
benlovell: This really feels like it should be somewhere else. What do you think?

app/domain/output_document.rb
```diff
@@ -0,0 +1,20 @@
+class OutputDocument
+  attr_accessor :appointment_summary, :variant
```
benlovell: Could these be reduced to readonly?

app/domain/output_document/html_renderer.rb
```diff
@@ -0,0 +1,23 @@
+class OutputDocument
+  class HTMLRenderer
+    include ActionView::Helpers::TextHelper
+
+    attr_accessor :output_document
```
benlovell: Same here, could this be readonly and / or private?

config/routes.rb
```diff
@@ -21,6 +21,12 @@
 
     resources :locations, only: [:index, :show]
 
+    resources :appointment_summaries, only: %i(new create), path: 'summary-document' do
+      post :view, on: :collection
+      post :download, on: :collection
+      post :print, on: :collection
+    end
```
dwhenry: Would the below be nicer - allows us tot use show instead of view - using resource instead of resources?

```
resource :appointment_summaries, only: %i(new create show), path: 'summary-document' do
  post :download, on: :member
  post :print, on: :member
end
```

app/models/appointment_summary.rb
```diff
@@ -0,0 +1,30 @@
...
+                :supplementary_defined_benefit_pensions, :appointment_type
+
+  %i(supplementary_benefits supplementary_debt
+     supplementary_ill_health supplementary_defined_benefit_pensions).each do |attribute_writer|
+    define_method("#{attribute_writer}=") do |value|
+      boolean = [true, 'true', '1', 1].member?(value)
+      instance_variable_set("@#{attribute_writer}", boolean)
+    end
+
+    alias_method(:"#{attribute_writer}?", attribute_writer)
```
benlovell: This could be a source of weird bugs. Perhaps we could instead define the predicate while iterating over the symbol list using the `[true, 'true', '1', 1].member?(value)` logic and just refer to that in the following writer method that gets defined?

config/routes.rb
```diff
@@ -21,6 +21,12 @@
 
     resources :locations, only: [:index, :show]
 
+    resources :appointment_summaries, only: %i(new create), path: 'summary-document' do
+      post :view, on: :collection
```
benlovell: You could nest these three `post` routes inside a `collection` block instead, then we can drop the `on` arguments for all three.

features/step_definitions/appointment_summary_steps.rb
```diff
@@ -0,0 +1,91 @@
...
+
+When(/^I view the appointment summary$/) do
+  generator = @page.generator
+  generator.submit_button.click
+
+  @page = Pages::AppointmentSummary.new
+end
+
+Given(/^I generate a valid appointment summary$/) do
+  step 'I visit the digital appointment summary generator'
```
benlovell: As a rule it's usually best to avoid nested steps at the cost of duplication. You could potentially either use `background`, duplicate those steps or just delegate them to actual methods instead. Breaking this things up later on becomes real tricky when steps call steps call steps.

app/controllers/appointment_summaries_controller.rb
benbarnett: Where would make sense, a view helper?

app/domain/output_document.rb
benbarnett: Done

app/domain/output_document/html_renderer.rb
benbarnett: Done

app/controllers/appointment_summaries_controller.rb
benbarnett: Have moved this to a helper, which cleans up the controller quite a bit. WDYT?

app/models/appointment_summary.rb
benbarnett: Good idea, have moved into a constant and used where applicable

app/views/appointment_summaries/create.html.erb
```diff
@@ -0,0 +1,95 @@
...
+            <%= link_to text, "##{id}", class: 't-aside--heading' %>
+          </h2>
+        <% end %>
+
+        <ul class="sticky-sidebar__list nav">
+          <% headers(2).each do |id, text| %>
+            <li>
+              <%= link_to text, "##{id}",
+                          class: [
+                              't-aside--link',
```
benlovell: These two lines are indented at four spaces.

PR #518, opened 2016-04-11
Add `autoprefixer-rails`

benbarnett:
As we write more complex components, bringing this in will simplify the SCSS we need to write.

https://github.com/postcss/autoprefixer


dwhenry:
:+1

dwhenry:
👍

PR #519, opened 2016-04-11
Improve screen reader labelling for currency fields

benbarnett:
Ensure screen readers will read out both the label and currency symbol for these fields.


app/views/calculators/_form_input_currency.html.erb
```diff
@@ -1,16 +1,17 @@
 <%
   error_messages = local_assigns[:error_messages] || []
   hint = local_assigns[:hint]
-  labelled_by = "#{field}-a"
+  labelled_by_symbol = "#{field}-symbol"
```
dwhenry: might be clearer to call this `labelled_by_currency`

PR #530, opened 2016-04-29
Update homepage text

alexkouz:
Section under 'Talk about your pension options' changed to reduce duplication. Also added 'printed' so that user knows summary won't just be verbal.


alexkouz:
👍

PR #536, opened 2016-05-09
Move `/summary-document` to `/after-your-appointment`

benlovell:
We had problems distinguishing between the 'start' page for appointment
summaries and the RESTful routes used for the various appointment
summary actions. In renaming the 'start' page to
`/after-your-appointment` we can remove the ambiguity and track the
various triggers for GTM easier.


dwhenry:
🚢

PR #538, opened 2016-05-17
Replace Draper decorators with POROs

benlovell:
Drops the Draper dependency in favour of PORO decorators. Draper seems pretty
dormant and likely won't be supporting the imminent Rails 5 release any time
soon.

It seems most of our decorators favour `SimpleDelegator` over Draper too, so
this also brings some consistencies.

### DIFF
--- a/app/decorators/guide_decorator.rb
+++ b/app/decorators/guide_decorator.rb
@@ -1,6 +1,19 @@
-class GuideDecorator < Draper::Decorator
-  delegate :id, :slug, :concise_label, :description,
-           :option?, :related_to_appointments?, :related_to_booking?
+class GuideDecorator
+  delegate(
+    :id,
+    :slug,
+    :concise_label,
+    :description,
+    :option?,
+    :related_to_appointments?,
+    :related_to_booking?,
+    :==,
+    to: :guide
+  )
+
+  def initialize(guide)
+    @guide = guide
+  end
 
   def url
     "/#{slug}"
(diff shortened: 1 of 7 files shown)


app/decorators/guide_decorator.rb
```diff
@@ -11,7 +24,7 @@ def title
   end
 
   def label
-    object.label.blank? ? title : object.label
+    guide.label.blank? ? title : guide.label
```
dwhenry:
```
guide.label.presence || title
```

app/decorators/guide_decorator.rb
```diff
@@ -1,6 +1,19 @@
...
+    :id,
+    :slug,
+    :concise_label,
+    :description,
+    :option?,
+    :related_to_appointments?,
+    :related_to_booking?,
+    :==,
+    to: :guide
+  )
```
dwhenry: Would subclassing `SimpleDelegator` be more concise than delegating all these methods?

benlovell: It was considered, but I decided against. Adding `SimpleDelegator` in the inheritance chain, with this being the base decorator for the govspeak and HTML decorators would mean this family of decorators forward absolutely everything. Also bearing in mind I've only added one more target: `:==` and generally I favour the explicitness of knowing right up front exactly which calls are forwarded - as is afforded by the `delegate` call.

app/decorators/guide_decorator.rb
benlovell: Good shout. Will do.

dwhenry:
🏃 🚢

PR #547, opened 2016-06-08
Slider on take cash in chunks calculator

benbarnett:
Users can adjust the amount of their first chunk using a slider

### DIFF
--- a/app/views/calculators/take_cash_in_chunks/_estimate.html.erb
+++ b/app/views/calculators/take_cash_in_chunks/_estimate.html.erb
@@ -1,22 +1,38 @@
+<div id="js-estimate">
+  <p>
+    If you take <span class="t-calculator-chunk"><%= number_to_currency(estimate.chunk) %></span> from your
+    <span class="t-calculator-pot"><%= number_to_currency(estimate.pot) %></span> pot, you could get:
+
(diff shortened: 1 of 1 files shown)


app/views/calculators/take_cash_in_chunks/_estimate.html.erb
```diff
@@ -1,22 +1,44 @@
...
-  <strong class="calculator__estimate__number t-calculator-pot-remaining">
-    <%= number_to_currency(estimate.pot_remaining) %> in your pot
-  </strong>
+<p>
+  <label for="slider">or try a different cash chunk:</label>
+  <%
+  max = estimate.chunk.to_f * Calculators::ARBITRARY_MULTIPLIER
+  if max >= estimate.pot.to_f
+    max = estimate.pot.to_i - 1
+  end
```
dwhenry: feel this calculation does not belong in the view, either in the estimate class or maybe a helper, this would also allow the constant to sit next to the calculation.

benbarnett: This actually made me realise we don't need the calculation anyway – we can just have the upper limit of the slider to just below pot size.

...which I have just done.

PR #548, opened 2016-06-08
Refresh calculator results when slider text input is changed

benbarnett:
Previously users had to hit enter or click outside the text field after changing the value of the slider. This PR adds test coverage and ensures that the form is updated on key change.


benlovell:
Nice. 🚀

PR #550, opened 2016-06-09
Visually highlight numbers on change

benbarnett:
* Visually highlight numbers on the Take cash in chunks and Leave your pot untouched calculators
* Provide a fallback slider maximum when a zero monthly contribution is provided

### DIFF
--- a/app/helpers/calculators_helper.rb
+++ b/app/helpers/calculators_helper.rb
@@ -1,0 +1,10 @@
+module CalculatorsHelper
+  CONTRIBUTION_MULTIPLIER = 5
+  DEFAULT_CONTRIBUTION_UPPER_LIMIT = 500
+
+  def leave_pot_untouched_slider_max(contribution)
+    max = contribution * CONTRIBUTION_MULTIPLIER
+    max = DEFAULT_CONTRIBUTION_UPPER_LIMIT if max.zero?
+    max
+  end
+end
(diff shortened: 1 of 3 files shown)


app/helpers/calculators_helper.rb
```diff
@@ -0,0 +1,7 @@
+module CalculatorsHelper
+  def leave_pot_untouched_slider_max(contribution)
+    max = contribution * 5
```
benlovell: What does the `5` represent? Might help to slap it in a constant.

app/helpers/calculators_helper.rb
```diff
@@ -0,0 +1,7 @@
...
+    max = 500 if max.zero?
+    max
```
dwhenry: maybe write this as a ternary instead?

benbarnett: Personally feel this would result in a longer less readable line but open to opinions, @benlovell any thoughts?

benlovell: I was going to say the same, but agree with @benbarnett - I think this is clearer than a ternary.

app/helpers/calculators_helper.rb
benbarnett: Have moved both the numbers here to constants

benlovell:
🚢 🐑 :shipit:

PR #551, opened 2016-06-09
Cap income lasting to 120 on adjustable income calculator

benbarnett:
The reality is drawdown can last for many years, or even
infinitely. But as this calculator is to illustrate how this
option works, we limit this number to an age of 120. Previously,
we showed (for example) "your income could last until you're 161"
– which can seem confusing.


benlovell:
🐑 🚀

PR #555, opened 2016-06-10
Added "how receiving a cash lump sum can affect your benefits"

gordon-cousins-pw:


benbarnett:
Hi @gordon-cousins-pw these are pull requests into the live site, is this what you intended or did you mean to put these into the prototype?

PR #562, opened 2016-06-13
Next and previous buttons

dwhenry:
The next and previous links have been added to the bottom of the calculator pages.

The implementation would allow the buttons to be added to any guide page
as it uses the Taxonomy class to determine the current guide pages position in 
the navigation tree.

### DIFF
--- a/app/helpers/guides_helper.rb
+++ b/app/helpers/guides_helper.rb
@@ -2,4 +2,24 @@ module GuidesHelper
   def book_an_appointment_link?
     !@guide.related_to_appointments?
   end
+
+  def previous_guide
+    return @previous_guide if defined?(@previous_guide)
+
+    previous_sibling = current_node.previous_sibling
+    @previous_guide = previous_sibling && GuideDecorator.cached_for(GuideRepository.new.find(previous_sibling.name))
+  end
+
+  def next_guide
+    return @next_guide if defined?(@next_guide)
+
+    next_sibling = current_node.next_sibling
+    @next_guide = next_sibling && GuideDecorator.cached_for(GuideRepository.new.find(next_sibling.name))
+  end
+
+  private
+
+  def current_node
+    @current_node ||= Taxonomy.instance.tree.detect { |node| node.name == @guide.id }
+  end
 end
--- a/features/journey_navigation.feature
+++ b/features/journey_navigation.feature
@@ -1,0 +1,18 @@
+Feature: Journey navigation
+  As a visitor to the website considering the 6 options
+  I want to be able to easily move between pages
+  So that I can find the best option for me
+
+  Scenario Outline: Guide navigation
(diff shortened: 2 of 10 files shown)


features/next_and_previous_navigation.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Next and previous navigation
+  As a visitor to the website considering the 6 options
+  I want to be able to easily move between pages
+  So that I can find the best option for me
+
+  Scenario Outline: Guide navigationb
```
benbarnett: Typo

benbarnett:
Thinking outloud: I wonder if `next-and-previous` is the best name for a tag... I think tags should say what sort of content it is, and then the site should decide how to render it. Whereas how it is now kind of has the content dictating it's layout.

I'm not sure of the solution though as it's likely other pages may want this too, and this approach makes that easy.

app/helpers/guides_helper.rb
```diff
@@ -2,4 +2,24 @@ module GuidesHelper
   def book_an_appointment_link?
     !@guide.related_to_appointments?
   end
+
+  def previous_guide
+    return @previous_guide if defined?(@previous_guide)
```
benlovell: Would this work as 

```ruby
@previous_guide ||= begin
 # ...
end
```

Same below with `next_guide`

content/adjustable_income.md
```diff
@@ -3,6 +3,7 @@ label: Get an adjustable income
 description: Your money is invested in a flexi-access drawdown fund to give you a regular income with the option to take cash when you need to.
 tags:
   - option
+  - next-and-previous
```
benlovell: Kinda feels to me like we're starting to misuse tags. Maybe this should be promoted to an actual front matter key like `paginates: true` or something similar?

app/helpers/guides_helper.rb
dwhenry: That would cause the code to be run twice for the nil option, while this approach avoids that duplication.

benlovell: I'm not sure I'm following, sorry. Can you explain what you mean by 'nil option'?

benlovell: Actually scratch what I just said. I see what you mean. Although I think the added clarity is worth the extra passes through these methods, personally. WDYT?

features/journey_navigation.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Journey navigation
+  As a visitor to the website considering the 6 options
+  I want to be able to easily move between pages
+  So that I can find the best option for me
+
+  Scenario Outline: Guide navigationb
```
benbarnett: Guide navigation**b**

benbarnett:
🚀

PR #565, opened 2016-06-14
Configurable online booking locations

benlovell:
This acts as a dumb feature flag for now. When the
`online_booking_location_uids` are specified, the corresponding locations are
enabled for online bookings and the start button is displayed to the customer.

### DIFF
--- a/features/customer_booking_request.feature
+++ b/features/customer_booking_request.feature
@@ -7,3 +7,8 @@ Scenario: Customer browses an online booking enabled location
   Given a location enabled for online booking
   When I browse for the location
   Then I can book online
+
+Scenario: Customer browses a regular location
+  Given no locations are enabled for online booking
+  When I browse for the location
+  Then I cannot book online
--- a/features/customer_booking_request.feature
+++ b/features/customer_booking_request.feature
@@ -1,0 +1,9 @@
+Feature: Customer creates a Booking Request
+  As a customer
+  I want to request an appointment
+  So I can understand my pension options
+
+Scenario: Customer browses an online booking enabled location
+  Given a location enabled for online booking
+  When I browse for the location
+  Then I can book online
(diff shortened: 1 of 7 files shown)


dwhenry:
I assume you considered using the location register to store the online booking enabled flag instead of adding it as a separate concern here and chose again it?

benlovell:
@dwhenry I did. It was my intention to not modify the existing `/locations.json` end point and associated plumbing over here. Obviously an additive change to the API _should_ have little to no impact, but if we're making changes over there every time we need something bookings related, this could quickly get out of hand.

Our intention once the bookings stuff stabilises is to rework consumers of the existing `/locations.json` API to accommodate the newer API, although that would of course require some additional behaviour to support such. Anyway, all thoughts for the future.

features/customer_booking_request.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Customer creates a Booking Request
+  As a customer
+  I want to request an appointment
+  So that I can meet for guidance
```
benbarnett: As these will be for phone appointments too, perhaps just `So I can find out more about my pension options` or something?

benbarnett:
🍬 as a 🌰

features/customer_booking_request.feature
benlovell: @benbarnett Changed to `So I can understand my pension options` which reads nicer too.

features/step_definitions/customer_booking_request_steps.rb
```diff
@@ -0,0 +1,31 @@
+Given(/^a location enabled for online booking$/) do
+  Locations.online_booking_location_uids = %w(ac7112c3-e3cf-45cd-a8ff-9ba827b8e7ef)
```
dwhenry: This alters the state of the system and could affect future tests. What are your thoughts on using the following instead
```
allow(Locations).to receive(:online_booking?).with('ac7112c3-e3cf-45cd-a8ff-9ba827b8e7ef').and_return(true)
```

benlovell: Good observation. I'm OK with being explicit about this for now. I'm providing the alternative in the `no locations are enabled for online booking` so we're at least covered in the places this will affect for now. If it becomes untenable in the future we'll address it then.

dwhenry:
🚢

PR #566, opened 2016-06-15
Remove sidebar navigation on location page

benbarnett:
We removed all the sidebar navigation a while back, but
this bit must have been missed. It was also pointing some pages
that no longer exist.


benlovell:
🔍 👍

PR #567, opened 2016-06-16
Online Booking Requests

benlovell:
Provides booking request functionality and interaction with both the booking locations API and the booking requests API.

Things to be aware of during review (and things that will be addressed in subsequent stories): 

1. No checks for error conditions or failures from the bookings API
2. APIs are called synchronously and reads are not cached

In development mode, stubbed API handlers are used to provide a success path through the bookings process. I'll likely change this in future to check for the presence of an environment variable and stub as necessary.

This is likely best reviewed on a commit-by-commit basis.

screen shot 2016-07-03 at 15 34 39


features/customer_booking_request.feature
```diff
@@ -12,3 +12,13 @@ Scenario: Customer browses a regular location
   Given no locations are enabled for online booking
   When I browse for the location
   Then I cannot book online
+
+@vcr
+Scenario: Customer makes an online Booking Request
+  Given a location enabled for online booking
```
benbarnett: a location `is` enabled?

benlovell: I think both ways read OK...

benbarnett: Not sure I feel that strongly! But equally I don't see how it makes sense as it is currently? 🤔

benlovell: Maybe it needs an Oxford comma. 😄

config/routes.rb
```diff
@@ -19,7 +19,12 @@
     get 'take-cash-in-chunks/estimate', to: 'calculators/take_cash_in_chunks#show'
     get 'take-whole-pot/estimate', to: 'calculators/take_whole_pot#show'
 
-    resources :locations, only: [:index, :show]
+    resources :locations, only: [:index, :show] do
+      member do
+        get '/booking_request/step_one',  to: 'booking_requests#step_one'
+        post '/booking_request/step_two', to: 'booking_requests#step_two'
```
benbarnett: Can we have the URLs with hyphens rather than underscores?

benbarnett: (to fit with the convention the rest of the site uses)

app/views/booking_requests/step_one.html.erb
```diff
@@ -0,0 +1,172 @@
+<div class="l-column-two-thirds">
+  <h1>Book an appointment at LOCATION NAME</h1>
```
benbarnett: Can this name be populated?

benlovell: Yeah, I just noticed that now. Whoops.

app/forms/booking_request_form.rb
```diff
@@ -6,6 +6,24 @@ class BookingRequestForm
...
+    step_one.validates :tertiary_slot, presence: true
+  end
+
+  with_options if: :step_two? do |step_two|
+    step_two.validates :first_name, presence: true
+    step_two.validates :last_name, presence: true
+    step_two.validates :email, presence: true, format: { with: /.+@.+\..+/ }
+    step_two.validates :telephone_number, presence: true
+    step_two.validates :memorable_word, presence: true
+    step_two.validates :appointment_type, inclusion: { in: %w(50-54 55-plus) }
```
benlovell: I'm going to change the flow so when this or the `dc_pot` validation is triggered we redirect to the eligibility screen, rather than presenting them as validation errors.

benlovell:
I've a bunch of content changes to make and I need to add the slots chosen from step 1 to be displayed in step 2, per the prototype and updated content.

app/forms/booking_request_form.rb
benbarnett: It looks like this has been done right?

config/locales/en.yml
```diff
@@ -18,12 +18,25 @@ en:
         strip_insignificant_zeros: true
 
   activemodel:
+    attributes:
+      booking_request_form:
+        primary_slot: Slot 1
+        secondary_slot: Slot 2
+        tertiary_slot: Slot 3
+        appointment_type: Age range
+        dc_pot: Direct contribution pot
```
benbarnett: This could be "defined contribution pot"

benbarnett: err *should* not *could*

app/forms/booking_request_form.rb
benlovell: Yep! 0ac168b

config/locales/en.yml
benlovell: Whoops. Will change now. Ta.

benbarnett:
👍 ^ ∞

PR #572, opened 2016-07-05
Use 2016/2017 basic rate income tax threshold

benbarnett:
Per the tax bands on https://www.gov.uk/government/publications/rates-and-allowances-income-tax/income-tax-rates-and-allowances-current-and-past

cc @amansinghb


benlovell:
Looks legit to me. Should we pass this around internally for review?

benbarnett:
Yeah I think @amansinghb was just checking this is definitely the right threshold as there seems to be some conflicting numbers

amansinghb:
After confirming with David Reid we should be using the following rates and thresholds for our calculators:

Personal Allowance - £11,000
Basic Rate - 20% - Up to 32,000
Higer Rate - 40% - £32,001 to £150,000
Additional Rate - 45% - Over £150,001

Would it be possible to make this change in a branch so we can test some figures on the calculator before pushing live?

@benbarnett @benlovell

benlovell:
I just amended @benbarnett's commit with the updated rates.

amansinghb:
Just tested it with David Reid from policy and the figures being calculated are correct. 👍

PR #575, opened 2016-07-12
Online booking customer details page tweaks

benbarnett:
* Allow validation summary to stretch to full width of container
* Persist radio button values on error state
* Rephrase `dc_pot` question as Yes, no, not sure. Yes and not sure are valid.

@benlovell that change to `dc_pot` might need some thinking on the [planner](https://github.com/guidance-guarantee-programme/planner/blob/master/spec/requests/create_a_booking_request_spec.rb#L74) side of things. Ideally we would store the exact answer the user provided, but I suppose we could coerce to `true` if easier?

### DIFF
--- a/spec/lib/booking_requests/api_mapper_spec.rb
+++ b/spec/lib/booking_requests/api_mapper_spec.rb
@@ -14,30 +14,46 @@ RSpec.describe BookingRequests::ApiMapper do
       appointment_type: '55-plus',
       accessibility_requirements: false,
       opt_in: false,
-      dc_pot: true
+      dc_pot: 'yes'
     )
   end
 
-  subject { described_class.map(booking_request) }
+  context 'mapping the booking_request form' do
+    subject { described_class.map(booking_request) }
 
-  it 'maps from the `BookingRequestForm` to requisite payload' do
-    expect(subject).to eq(
-      booking_request: {
-        location_id: location_id,
-        name: 'Lucius Needful',
-        email: 'lucius@example.com',
-        phone: '0208 244 3987',
-        memorable_word: 'meseeks',
-        age_range: '55-plus',
-        accessibility_requirements: false,
-        marketing_opt_in: false,
-        defined_contribution_pot: true,
-        slots: [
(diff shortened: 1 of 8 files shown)


benbarnett:
@benlovell after a quick discussion here, it seems fine to just coerce as `true` if the value is `yes` or `not-sure`. I'll take a look and see what I can do..!

benlovell:
It's a kinda weird setup but remember that in practice those fields would never be displayed in their error state, since if they're included in the errors (`dc_pot` and `appointment_type`) the customer is redirected to the ineligibility page.

benbarnett:
They are displayed if you provide an invalid email address though – say you've provided eligible answers for `dc_pot` and `appointment_type` but missed a couple of the other fields. That's what this change is catering for.

benlovell:
Cool. 👍

benbarnett:
In terms of the `yes` and `not-sure` == `true` coercion, do you think that should happen at this side? So we just post a boolean to the API?

benlovell:
Yeah, makes sense to do that work over here and pass as true / false. Although then we'd have no record whether they said they were unsure, if that matters then we'll need to change the API to accommodate. Is that something we need to care about?

benbarnett:
Yeah exactly, it sounds like that would be acceptable at this stage but let me find out for sure!

spec/lib/booking_requests/api_mapper_spec.rb
```diff
@@ -18,26 +18,42 @@
...
+            { priority: 3, date: '2016-01-02', from: '0900', to: '1300' }
+          ]
+        }
+      )
+    end
+  end
+
+  context 'coerces the `dc_pot` strings to booleans' do
+    it 'coerces `yes` to true' do
+      expect(described_class.dc_pot_as_boolean('yes')).to be_truthy
```
benlovell: I think we should use the more specific `be_true` / `be_false` matchers here and in the specs below, as we're explicitly testing it has been coerced to a boolean.

benlovell:
LGTM ⛵️🏁

PR #576, opened 2016-07-14
Include `booking_location_id` in JSON payload

benlovell:
The planner API for booking requests expects this now.

The rationale for this is we can make more efficient lookups for location
sub-trees in the planner frontend, when querying for location-specific
booking requests. This would be in the context of a booking's manager who is
only permitted to view bookings within their own location sub-tree.


benbarnett:
So nice

PR #577, opened 2016-07-15
Allow 12 factor configuration for location IDs

benlovell:
Allows us to configure pilot booking location IDs using a comma
separated list of IDs in the `BOOKING_LOCATION_UIDS` environment
variable. Will safely evaluate to an empty array when unset.


benlovell:
Hold fire on this, I think it warrants a spec or two.

benlovell:
This doesn't make sense once we have lots of locations and their sub-locations.

PR #578, opened 2016-07-15
Configure staging online booking locations

benlovell:
Enable online booking for Hackney and its associated locations in the
staging environment.


benbarnett:
Cool 👍

PR #579, opened 2016-07-15
Bump `booking_locations` gem to send `Accept`

benlovell:
Otherwise GDS-SSO incorrectly identifies the type of request, and doesn't
use the bearer token when authenticating.

Ensure the `Accept` header is passed as JSON

Again, this ensures GDS-SSO correctly identifies the request as API in nature,
and does the bearer token authentication - not the usual oauth2 dance.


benbarnett:
👍

PR #580, opened 2016-07-19
Online booking tweaks

benbarnett:
* Minor copy tweaks
* Only show ineligibility page if all other errors are satisfied

### DIFF
--- a/app/controllers/booking_requests_controller.rb
+++ b/app/controllers/booking_requests_controller.rb
@@ -14,12 +14,12 @@ class BookingRequestsController < ApplicationController
   def complete
     @booking_request = BookingRequestForm.new(location_id, booking_request_params)
 
-    if @booking_request.step_two_valid?
+    if @booking_request.step_two_valid_excluding_eligibility?
+      return redirect_to booking_request_ineligible_location_path(id: location_id) unless @booking_request.eligible?
+
       BookingRequests.create(@booking_request)
       redirect_to booking_request_completed_location_path(id: location_id)
     else
-      return redirect_to booking_request_ineligible_location_path(id: location_id) unless @booking_request.eligible?
-
       render :step_two
     end
   end
--- a/app/forms/booking_request_form.rb
+++ b/app/forms/booking_request_form.rb
@@ -20,7 +20,7 @@ class BookingRequestForm
     step_two.validates :memorable_word, presence: true
     step_two.validates :appointment_type, inclusion: { in: %w(50-54 55-plus) }
     step_two.validates :accessibility_requirements, inclusion: { in: %w(0 1) }
-    step_two.validates :opt_in, inclusion: { in: %w(0 1) }
+    step_two.validates :opt_in, inclusion: { in: %w(1) }
     step_two.validates :dc_pot, inclusion: { in: %w(yes not-sure) }
   end
 
--- a/app/forms/booking_request_form.rb
+++ b/app/forms/booking_request_form.rb
@@ -53,6 +53,13 @@ class BookingRequestForm
     valid?
   end
 
+  def step_two_valid_excluding_eligibility?
+    @step = 2
+    valid?
+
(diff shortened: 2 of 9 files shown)


app/controllers/booking_requests_controller.rb
```diff
@@ -14,12 +14,12 @@ def step_two
...
+    if @booking_request.step_two_valid_excluding_eligibility?
+      return redirect_to booking_request_ineligible_location_path(id: location_id) unless @booking_request.eligible?
+
       BookingRequests.create(@booking_request)
       redirect_to booking_request_completed_location_path(id: location_id)
     else
-      return redirect_to booking_request_ineligible_location_path(id: location_id) unless @booking_request.eligible?
-
       render :step_two
     end
```
dwhenry: Don't feel like this change improved readability.

Have you consider:
```
if @booking_request.step_two_valid? && @booking_request.eligible?
  # do A
elsif @booking_request.step_two_valid? # i.e. ineligible
  # do B
else
  # do C
end
```

The reason I suggest this is that step 2 is complete (i.e. valid) regardless of eligibility criteria.

app/forms/booking_request_form.rb
```diff
@@ -53,6 +53,13 @@ def step_two_valid?
     valid?
   end
 
+  def step_two_valid_excluding_eligibility?
+    @step = 2
+    valid?
+
+    (errors.keys - %i(appointment_type dc_pot)).empty?
```
dwhenry: if we are going to do this can we extract `%i(appointment_type dc_pot)` into a constant.

benlovell: Agreed, obviously replacing the same in `eligible?` too.

app/forms/booking_request_form.rb
```diff
@@ -20,7 +20,7 @@ class BookingRequestForm
     step_two.validates :memorable_word, presence: true
     step_two.validates :appointment_type, inclusion: { in: %w(50-54 55-plus) }
     step_two.validates :accessibility_requirements, inclusion: { in: %w(0 1) }
-    step_two.validates :opt_in, inclusion: { in: %w(0 1) }
```
benlovell: What happened here? Doesn't this mean they *must* opt in now?

app/views/booking_requests/step_two.html.erb
```diff
@@ -139,7 +139,7 @@
           <% if @booking_request.errors.include?(:opt_in) %>
             <span class="error-message"><%= @booking_request.errors[:opt_in].to_sentence %></span>
           <% end %>
-          <%= f.check_box :opt_in, class: 't-opt-in' %> Yes, I would like to receive marketing information
+          <%= f.check_box :opt_in, class: 't-opt-in' %> I accept the <a href="/privacy" target="_blank">terms and conditions</a>
```
benlovell: Ah I see now. Ignore the previous comment regarding validation.

app/forms/booking_request_form.rb
benlovell: Edit: Ignore this, I realised the meaning below.

app/controllers/booking_requests_controller.rb
benlovell: I don't think that improves readability either. Perhaps `eligible?` needs to operate like a separate step, following the existing pattern in that form object? I think that would be cleaner. Also we need specs for these behaviour changes as they're fundamental to this process.

benlovell:
I feel like the change to the meaning of `opt_in` should be documented better in the commit message. It might lend some meaning to the naming of `opt_in` at least.

benbarnett:
As just quickly discussed I think we're going to leave the eligibility as they are (were) for now. Will close this PR and re-raise with the copy amends separately 😅

benlovell:
*Tis better to have loved and lost, than never to have loved at all*

PR #581, opened 2016-07-19
Use opt_in as terms and conditions acceptance

benbarnett:
This field is now required, and is a reference to the customer
accepting Pension Wise's terms and conditions.


benbarnett:
@benlovell have updated the specs on this one now

PR #583, opened 2016-07-27
Update pension scams page

alexkouz:
Link to PDF of example scams updated after request from The Pensions Regulator. PDF now 5 pages.


alexkouz:
@benlovell The old PDF was 1mb. Can you tell if this has changed now it's down to 5 pages? Ta.

benbarnett:
@alexkouz it's about ~760kb now. Do we round to the nearest mb?

alexkouz:
@benbarnett Sure, why not.

benbarnett:
🔨

PR #584, opened 2016-07-27
Improve online booking journey for users without javascript

benbarnett:
The calendar is progressively enhanced up from three standard
select boxes, so we simply fall back to these for users where
javascript is not available.


benlovell:
Niceness. 💯

PR #585, opened 2016-07-28
Digital Summary doc start page content change

amansinghb:
Removed text from the digital summary document start page that states
customer will get a printed version as well.


benbarnett:
👍

PR #587, opened 2016-07-29
Cache booking locations in memory

benlovell:
Use a read-through cache to store booking locations JSON in memory and
save a bunch of roundtrips to the API during the booking process. The
locations have a TTL of two hours by default.

For the timebeing we're configured to use the memory store cache, but
it may make more sense in future to use the redis cache or something
shared, not per-dyno.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -58,7 +58,7 @@ Rails.application.configure do
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :memory_store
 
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -58,7 +58,7 @@
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :memory_store
```
dwhenry:
```
config.cache_store = :redis
```
?

benlovell: Read the commit message 😉

benlovell:
I think I'll extract this to the `booking_locations` gem in the near future.

benbarnett:
Sumptuous 👍

PR #588, opened 2016-07-29
Configure HTTP connections for bookings API

benlovell:
Bumps the default timeout to 10 seconds for both connect and read.
Also, since the endpoint is not considered idempotent I have disabled
retries.

If the timeouts persist, I'll make the API call asynchronously and handle it
that way. There is no overriding reason why this interaction couldn't be
backgrounded anyway, as we don't present the response to the user as the
request is always considered valid.

### DIFF
--- a/lib/booking_requests/api.rb
+++ b/lib/booking_requests/api.rb
@@ -9,10 +9,30 @@ module BookingRequests
     private
 
     def connection
-      HTTPConnectionFactory.build(api_uri).tap do |c|
+      HTTPConnectionFactory.build(api_uri, connection_options).tap do |c|
         c.headers[:accept] = 'application/json'
         c.authorization :Bearer, bearer_token if bearer_token
       end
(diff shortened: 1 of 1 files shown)


benbarnett:
👍

dwhenry:
Are we expecting this change to remove the timeouts or to hide them is they respond in 5-10 seconds instead of less than 5? 

My feeling is that an appropriate timeout should be smaller rather than larger given directly affects the response time. 

If response time/connectivity is an issue shouldn't we extract this to a job now?

benlovell:
This is important context that I failed to add in the original PR:

This is entirely down to heroku idling the staging dynos due to lack of use. When they're running, the responses return in the range of milliseconds. My view is that we'll reduce the timeouts to perhaps a second or two when we move to production. Obviously these are 12-factor configured, as you can see in the diff.

PR #589, opened 2016-07-29
Update pension protection guide

alexkouz:
FCA wasn't spelled out.

### DIFF
--- a/content/protection.md
+++ b/content/protection.md
@@ -13,9 +13,9 @@ If your employer goes out of business you won’t lose your pension pot. This is
 
 ## If your provider can’t pay
 
-You can be compensated for up to 100% of the value of your pot if your pension provider can’t pay you and is authorised by the FCA.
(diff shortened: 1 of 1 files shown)


content/protection.md
```diff
@@ -13,7 +13,7 @@ If your employer goes out of business you won’t lose your pension pot. This is
 
 ## If your provider can’t pay
 
-You can be compensated for up to 100% of the value of your pot if your pension provider can’t pay you and is authorised by the FCA.
+You can be compensated for up to 100% of the value of your pot if your pension provider can’t pay you and is authorised by the Financial Conduct Authority (FCA).
```
benlovell: @benbarnett Do we use `abbr` for this?

benbarnett: Yep, good spot. I can't remember if there's markdown syntax for this. Will check first thing Monday.

benbarnett: Thinking about this, we don't actually use `abbr` tags elsewhere on the site. Perhaps we can do revisit in a separate PR for all abbreviations. So this is cool as is.

PR #590, opened 2016-08-01
GTM tracking events for form validation errors

benbarnett:
Can be used to monitor how many users are seeing errors with
their input, and on which step the input error occurred.


benlovell:
Beaut. 🐑

PR #593, opened 2016-08-02
Enable Hackney and its associated locations

benlovell:
Enables Hackney and its associated booking centres for online bookings.


benbarnett:
👍

PR #595, opened 2016-08-04
Rework feedback for testability

benlovell:
* Rework controller actions to follow 'RESTful' structure
* Make Zendesk client pluggable / stubable
* Adds features to specify feedback behaviour

### DIFF
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -1,0 +1,54 @@
+class FeedbacksController < ApplicationController
+  layout 'full_width'
+
+  def new
+    @booking_feedback = BookingFeedbackForm.new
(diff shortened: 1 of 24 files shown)


app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  def render_html_response
+    if @booking_feedback.valid?
+      redirect_to thanks_feedback_path
+    else
+      render :new
+    end
+  end
+
+  def render_result(status)
+    render partial: 'result', status: status if request.xhr?
```
benbarnett: Is this guard still needed? Looks like this is only called from `render_xhr_response`

benbarnett:
💯

PR #596, opened 2016-08-04
Open feedback tickets from the customer

benlovell:
Reworks the ticket creation to make them appear as though they're sent
directly from the customer. This allows us to track customer
interactions in zendesk much more efficiently.


benbarnett:
Likey 👍

PR #597, opened 2016-08-04
Cache locations via booking locations gem

benlovell:
Location caching is now provided (and configurable) by the booking
locations gem since v0.6. We should use that instead of the customised
caching code in this project.


config/initializers/online_booking_locations.rb
```diff
@@ -13,11 +13,9 @@
...
-                         BookingLocations::StubApi.new
-                       else
-                         require 'cached_booking_locations_api'
-                         CachedBookingLocationsApi.new
-                       end
+if Rails.env.development?
+  require 'booking_locations/stub_api'
+  BookingLocations.api = BookingLocations::StubApi.new
+else
+  BookingLocations.cache = Rails.cache
```
benbarnett: Do we need to consider the test environment here? Or does that just get bypassed elsewhere?

benlovell: Test environment does the right thing explicitly, depending on the needs of the particular test. 👍

benlovell:
@benbarnett is this ok to merge?

benbarnett:
Yep 👍

PR #600, opened 2016-08-05
Annuities calculator

dwhenry:
Prototype implementation for testing

### DIFF
--- a/app/views/calculators/annuity/_calculator.html.erb
+++ b/app/views/calculators/annuity/_calculator.html.erb
@@ -1,0 +1,10 @@
+<%= render 'form', form: form %>
+<% if form.valid? %>
+  <%= render 'estimate',
+    estimate: form.estimate,
+    age: form.age.to_i,
+    pension_income: form.pension_income,
+    life_expectancy: form.life_expectancy.to_i,
+    valuation: form.estimate[:valuation]
+  %>
+<% end %>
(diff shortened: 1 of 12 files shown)


benbarnett:
Just a reminder to push the monthly -> annual change up when you can.

Think it just makes sense as we'll forget the context over the weekend. Well I will anyway.

dwhenry:
@benbarnett rename done. 😄

app/views/calculators/annuity/_calculator.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= render 'form', form: form %>
+<% if form.valid? %>
+  <%= render 'estimate',
+    estimate: form.estimate,
+    age: form.age.to_i,
+    pension_income: form.pension_income,
+    life_expectancy: form.life_expectancy.to_i,
+    valuation: form.estimate[:valuation]
```
benbarnett: We're passing the whole `estimate` to this partial – is it worth just referencing it from there?

dwhenry: this is following the established pattern. I can see some benefit - not having to have the long accessor where it is used multiple times.

But agree it most likely isn't worth it in this case, but would suggest we revisit it when we make this production ready.

benbarnett: 🆗 👌

benbarnett:
More than enough for a prototype! 👍

PR #601, opened 2016-08-08
Styleguide homepage tidyup

Guntrisoft:
Remove old header from styleguide homepage layout and replace with newer nav component and smaller PW logo


Guntrisoft:
@benbarnett

benbarnett:
Nice one. Want to squash these commits and perhaps reword so it doesn't run over to the description?

PR #602, opened 2016-08-09
Converting circles to use css instead of sprite

Guntrisoft:
The circles used at the moment on the site are a bit rough round the edges on a retina screen - I've converted them to use border-radius and got rid of the sprites that are used - less CSS, and better looking on almost every browser we care about.

On IE8 they end up as coloured squares, which matches the bullet points used, as they use border-radius too.

screen shot 2016-08-09 at 16 26 16


benbarnett:
Nice. There's some [steps in the package.json](https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/package.json#L22-L24) as well that could come out.

Guntrisoft:
Knew there was something i'd missed! - removed!

PR #606, opened 2016-08-12
Adding wraith to test pensionwise pages across device widths

Guntrisoft:
* run wraith: 'npm run wraith'

### DIFF
--- a/wraith/configs/capture.yaml
+++ b/wraith/configs/capture.yaml
@@ -1,0 +1,80 @@
+##############################################################
+##############################################################
+# This is an example configuration provided by Wraith.
+# Feel free to amend for your own requirements.
+# ---
+# This particular config is intended to demonstrate how
+# to use Wraith in 'capture' mode, which is best suited to
+# comparing a test and live version of the same website.
+#
+# `wraith capture capture.yaml`
+#
+##############################################################
+##############################################################
+
+# (required) The engine to run Wraith with. Examples: 'phantomjs', 'casperjs', 'slimerjs'
+browser: "phantomjs"
+
+# (required) The domains to take screenshots of.
+domains:
+  local: "http://localhost:5000"
+  live: "https://www.pensionwise.gov.uk"
+
+# # (required) The paths to capture. All paths should exist for both of the domains specified above.
+paths:
+  home: /
+  styleguide: /styleguide
+  styleguide_base: /styleguide/base
+  styleguide_govspeak: /styleguide/govspeak
+  styleguide_components: /styleguide/components
+  styleguide_govuk_elements: /styleguide/govuk_elements
+  pension_pot_options: /pension-pot-options
+  tax: /tax
+  appointments: /appointments
+  mix_options: /mix-options
+  browse_tax_and_getting_advice: /browse/tax-and-getting-advice
+  contact: /contact
+  summary_document_new: /summary-document/new
+  locations: /locations?postcode=
+  cookies: /cookies
+  feedback_new: /feedback/new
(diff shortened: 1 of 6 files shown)


wraith/configs/spider.yaml
```diff
@@ -0,0 +1,60 @@
+##############################################################
```
benbarnett: Do we need these example ones? Out of interest what is brought in by the gem, do we need to include all this stuff in the repo?

Guntrisoft: true - I'll tidy up some of the files which aren't used.

wraith/configs/capture.yaml
```diff
@@ -0,0 +1,78 @@
...
+browser: "phantomjs"
+
+# (required) The domains to take screenshots of.
+domains:
+  local:       "http://localhost:5000"
+  live:        "https://www.pensionwise.gov.uk"
+
+# # (required) The paths to capture. All paths should exist for both of the domains specified above.
+paths:
+  home:     /
```
benlovell: Errant spacing, wurd up.

wraith/configs/capture.yaml
```diff
@@ -0,0 +1,78 @@
...
+  styleguide:    /styleguide
```
benlovell: Hey I just met you, and this is crazy, but here's some whitespace, delete it maybe?

benbarnett:
Wonder if it's worth adding something to the readme about needing the server running on `5000`? The error is fairly opaque if not. Otherwise this is awesome. My laptop got a decent workout too! 👍

benlovell:
Yep, replicate this perhaps: https://github.com/guidance-guarantee-programme/pension_guidance/blob/master/config/puma.rb#L15

Guntrisoft:
@benlovell @benbarnett gone with a script which prompts for port number before running the test - wraith appears to only accept a config path rather than config items as arguments so I've gone down the route of creating a temporary config and replacing the port number into it.

benlovell:
Needs rebae.

wraith/bin/wraith.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+read -p "Specify the port number that your local dev environment is running on: " port
+
+port=${port:-5000}
```
benlovell: I don't think you need to even prompt for this. Check if `PORT` is set or default to `3000`.

benbarnett:
👑
👴🏻

✅

wraith/bin/wraith.sh
Guntrisoft: Do you mean something like `PORT=5000 npm run wraith` ?

benlovell: On reflection, why don't we just hard code port 5000 in the config file and dispense with all the trickery? I think it's fair to assume this project is bound to 5000 locally, since that's what foreman starts you up on.

Guntrisoft: This is what I had in the first place, but @benbarnett said the warnings were silent on the fact that it was running on a different port

benbarnett: Just been going through the wraith source code and it looks like it's a YAML config or nothing. Would be nice to specify this stuff with an environment variable but agree that the amount of trickery involved probably doesn't provide as much value. And given our readme recommends `foreman` anyway, this lines up.

So shall we just get rid?

benlovell: Makes sense to me. ✅

benlovell:
Lovely.

PR #607, opened 2016-08-15
Get Twilio number from locations.json

dwhenry:
This removed the dependency on the switchboard
application and simplifies the code for the location
search.


benlovell:
Let's make sure this gets a good work out in staging before promoting it. 👍

PR #608, opened 2016-08-15
Update to latest annuity market average rates

benbarnett:
There has been quite a significant change in annuity rates in
recent months. This change brings the guaranteed income
calculator inline with this.


benlovell:
👍

PR #609, opened 2016-08-15
Deploy using heroku auto-deploys

benlovell:
This deployment configuration was created before the existence of
heroku pipelines. Our other projects use pipelines and auto-deploys so
we should use those features here too.


benbarnett:
Hi there yes please 👍

PR #611, opened 2016-08-17
Upgrade govuk elements/toolkit/template

Guntrisoft:
This branch updates the following dependencies on the pension_guidance project:

* https://github.com/alphagov/govuk_template - to 0.18.0
* https://github.com/alphagov/govuk_frontend_toolkit to 4.14.1
* https://github.com/alphagov/govuk_elements to 1.2.2

PR #614, opened 2016-08-22
Adds homepage experiment to wraith config

benbarnett:
This page is a first class citizen as it is in front of customers,
so this will have wraith compare visual changes to it as well.

This was plucked from #611.


Guntrisoft:
👍

PR #615, opened 2016-08-25
Enable Barry / Cardiff and Vale online bookings

benlovell:
Enables Barry and its associated booking locations for online bookings.


benbarnett:
👍

PR #618, opened 2016-08-26
Add missing close div tag

dwhenry:
Adds a missing `</div>` at the end and fixes allignment

### DIFF
--- a/app/views/booking_requests/step_two.html.erb
+++ b/app/views/booking_requests/step_two.html.erb
@@ -157,41 +157,42 @@
     <p>Your chosen appointment dates are below.</p>
 
     <div class="SlotPicker-choices is-chosen SlotPicker--selected SlotPicker--read-only">
-     <div class="SlotPicker-choice is-chosen">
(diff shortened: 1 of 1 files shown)


app/views/booking_requests/step_two.html.erb
```diff
@@ -157,51 +157,52 @@
...
+
+    <p><%= link_to 'Change dates',
         booking_request_step_one_location_path(
-          id: location_id,
-          booking_request: {
-            primary_slot: @booking_request.primary_slot,
-            secondary_slot: @booking_request.secondary_slot,
-            tertiary_slot: @booking_request.tertiary_slot
-          }
+            id: location_id,
```
benlovell: This seems to have been auto-incorrected 😄

dwhenry: This is fixed now.. and think I have corrected the formatting rule that went wrong..

benlovell:
Nice catch. :shipit:

PR #622, opened 2016-09-01
Update email validation to stop ZenDesk errors

dwhenry:
This means emails can not contain spaces and
trailing spaces are removed inline with:

https://support.zendesk.com/hc/en-us/community/posts/203455266-Requester-email-is-already-used-by-another-user


benbarnett:
👍

PR #623, opened 2016-09-01
Swallow ZenDesk errors

dwhenry:
Stop errors from the ZenDesk API from being
reported to end-users. Error will be raised
to bugsnag and need to be manually handled.

### DIFF
--- a/lib/zen_desk/client.rb
+++ b/lib/zen_desk/client.rb
@@ -7,8 +7,14 @@ module ZenDesk
       @client = client
     end
 
+    def create_ticket(options)
+      create_ticket!(options)
+    rescue ZendeskAPI::Error::RecordInvalid => e
+      Bugsnag.notify(e)
+    end
+
     # rubocop:disable Metrics/ParameterLists
-    def create_ticket(name:, email:, message:, subject:, tags:, custom_fields: nil)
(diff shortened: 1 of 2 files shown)


dwhenry:
I'm not 100% this is the best location/way to do this do any thoughts welcome.

benlovell:
Was this resolved by #622 or is that a different issue?

dwhenry:
@benlovell #622 resolved the issues we had with the invalid email. 

This PR is designed to ensure the customer does not get the error page if new and exciting errors occur on the ZenDesk end that we do not know about/are not expecting.

As a result, we would need to manually create any failed tickets, however, I think this is better than showing an error page with no context.

benlovell:
I think it's too generalised in that case. The specific error being raised was due to a duplicate requester. That seems more domain-specific at our end, rather than just catching general errors. Perhaps we should figure out how to work around duplicates to deal with the specific case that was raised?

dwhenry:
> The specific error being raised was due to a duplicate requester

This is not the case, the error is related to space at the end of the email address (a have a webpage explaining this somewhere) #622 fixes that issue. The idea here is to catch unexpected ZenDesk errors rather than passing then to the end user.

PR #626, opened 2016-09-04
Exit with correct exit code

benlovell:
Returns the correct exit code, rather than raising an exception.


dwhenry:
Just reading up on `$CHILD_STATUS` and found the below: https://github.com/bbatsov/rubocop/issues/1747

Not sure if it is relevant?

benlovell:
@dwhenry it is relevant but not an issue here since I already `require 'English'`, which gives you the `$CHILD_STATUS` alias for `$?`.

dwhenry:
kool. 😄  🚢

PR #628, opened 2016-09-05
Stop /home-alternative from being indexed by search engines

Guntrisoft:


Guntrisoft:
@benbarnett

benlovell:
Perhaps we should just specify the canonical URL - inclusive of the scheme (HTTPS) - globally?

benbarnett:
Agree with @benlovell. Perhaps the `content_for` is to provide the path only, and the FQDN is just part of the template.

benlovell:
@Guntrisoft there's an initializer in `config/initializers/canonical_rails.rb` where you can configure this globally - in that case.

Guntrisoft:
@benlovell @benbarnett re-jigged the commit, thoughts?

benbarnett:
image

PR #630, opened 2016-09-06
Add transfer pension guide to navigation

benbarnett:
Replaces `combine_pots` guide, which this commit also removes.


benlovell:
Greatness.

PR #635, opened 2016-09-15
Slot picker tweaks from user research findings

benbarnett:
See individual commits for more context.

slot-picker


dwhenry:
LGTM.

PR #637, opened 2016-09-21
Updating home experiment stats figures

Guntrisoft:
from https://www.gov.uk/performance/pension-wise


Guntrisoft:
@benbarnett quick one - this is based on an email from John

benbarnett:
Looks good to me.

PR #641, opened 2016-09-22
Rely on gem-specified phantomjs binary

benlovell:
We should simply rely on the gem-specific phantomjs binary. Version
mismatches between the locked `1.9.8.0` and your local binary won't
cause an issue in development / test.


benlovell:
cc @Guntrisoft @benbarnett

benbarnett:
You know what? Sure.

PR #647, opened 2016-09-30
Surface booking location information in the data layer

benbarnett:
Analytics would like to see the booking centre and delivery centre
surfaced in the data layer so we can effectively track marketing
campaigns.


Guntrisoft:
👍

PR #657, opened 2016-10-11
Update summary document template

dwhenry:
- better images
- page numbers
- align logo on right of first page

__Pending testing__


app/domain/output_document.rb
```diff
@@ -16,4 +16,8 @@ def html
   def pdf
     Princely.new.pdf_from_string(html)
   end
+
+  def supplementary_pension_transfers
```
benlovell: Should / could this be a predicate?

PR #658, opened 2016-10-12
Book appointment link a/b test

dwhenry:
Add 'Book appointment link' experiment

Remove 'home-intro__link' class from book
appointment link so that it uses standard 
link colouring instead of being grey text.

Extract 'google experiments' JS to its own partial.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -5,6 +5,10 @@ class HomeController < ApplicationController
     expires_in Rails.application.config.cache_max_age, public: true
   end
 
+  def experiment
+    render 'experiment'
+  end
+
   def footer?
     false
   end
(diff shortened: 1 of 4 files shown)


app/views/testing/_user_zoom.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  _uzactions.push(['_setID', '<%= ENV['USER_ZOOM_ID'] %>']);
+  _uzactions.push(['_setSID', '<%= ENV['USER_ZOOM_SID'] %>']);
+  _uzactions.push(['_start']);
+
+  (function() {
+    var uz = document.createElement('script');
+    uz.type = 'text/javascript';
+    uz.async = true;
+    uz.charset = 'utf-8';
+    uz.src = 'https://cdn4.userzoom.com/trueintent/js/uz_til.js?cuid=C85E515772DAE311BEDA0022196C4538';
```
benbarnett: Does the `cuid` here need to be an env variable too?

dwhenry: it has never changed, but think it would be nice for it to be one as well.

PR #659, opened 2016-10-12
Refactor UserZoom JS include

dwhenry:
Extract UserZoom JS to it's own file and
set ID, SID and customer ID from ENV variables instead 
of hard-coding into the page.

### DIFF
--- a/app/views/layouts/_base.html.erb
+++ b/app/views/layouts/_base.html.erb
@@ -66,40 +66,7 @@
 <% content_for :body_end do %>
   <%= javascript_include_tag 'application' %>
 
-  <% if homepage_experiment_id %>
-    <%= javascript_include_tag 'https://cdn3.userzoom.com/uz.js?cuid=C85E515772DAE311BEDA0022196C4538' %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/_base.html.erb
```diff
@@ -66,40 +66,7 @@
...
-      uz.async = true;
-      uz.charset = 'utf-8';
-      uz.src = 'https://cdn4.userzoom.com/trueintent/js/uz_til.js?cuid=C85E515772DAE311BEDA0022196C4538';
-      var s = document.getElementsByTagName('script')[0];
-      s.parentNode.insertBefore(uz, s);
-    })();
-    //]]>
-    </script>
-  <% end %>
+  <%= render 'testing/user_zoom' %>
```
benbarnett: Tiny thing but shall we call this directory `research` instead of testing? Just feels a bit less ambiguous.

dwhenry: yes.. I like it. :)

benbarnett:
Nice one 👍

PR #661, opened 2016-10-12
Update statistics on homepage

benbarnett:
Per the [performance platform](https://www.gov.uk/performance/pension-wise)


Guntrisoft:
👍

PR #674, opened 2016-10-26
[DO NOT MERGE] Allow online booking to be enabled from Locator

dwhenry:
Use `online_booking_enabled` flag to determine
if location is eligible for online booking instead
of hardcoded list of UID's.

**Ensure all locations are correct set with online booking enabled before merging**


lib/locations/location.rb
```diff
@@ -1,3 +1,19 @@
 module Locations
-  Location = Struct.new(:id, :name, :address, :booking_location_id, :phone, :hours, :twilio_number, :lat_lng)
+  class Location
+    include ActiveModel::Model
```
benlovell: What's the rationale for this? If you just want to support kwarg initialisation maybe just convert the `Struct` to `OpenStruct`? Also for the predicate method you could just pass a block to define methods within the scope of the struct

```ruby
Struct.new(:id, :blah) do
  def online_booking_enabled?
    # ...
  end
end
```

spec/support/helpers/test_location.rb
```diff
@@ -0,0 +1,43 @@
+class TestLocation
+  def self.london # rubocop:disable Metrics/MethodLength
```
benlovell: May as well just move the rule disable pragma to the top of the file?

dwhenry:
@benlovell as discuss I am going to merge this as is. :)

benlovell:
@dwhenry Did you see my latest comment?

PR #677, opened 2016-11-04
Content changes to the online booking journey

amansinghb:
Content changes based on the latest lab session outcomes.

### DIFF
--- a/app/views/booking_requests/step_one.html.erb
+++ b/app/views/booking_requests/step_one.html.erb
@@ -50,6 +50,9 @@
         <%= f.button('Continue', type: 'submit', class: 'button t-continue') %>
       </p>
     <% end %>
+    <p>This service is in development.</p>
+
+    <p>You may not get one of the 3 dates you’ve requested. <br>If there’s a problem, we’ll call or email to confirm an appointment date with you.</p>
   </div>
 
   <div class="l-column-half">
--- a/app/views/booking_requests/step_two.html.erb
+++ b/app/views/booking_requests/step_two.html.erb
@@ -130,7 +130,7 @@
           <% if @booking_request.errors.include?(:accessibility_requirements) %>
             <span class="error-message"><%= @booking_request.errors[:accessibility_requirements].to_sentence %></span>
           <% end %>
-          <%= f.check_box :accessibility_requirements, class: 't-accessibility-requirements' %> I have accessibility needs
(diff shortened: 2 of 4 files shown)


app/views/booking_requests/step_one.html.erb
```diff
@@ -50,6 +50,9 @@
         <%= f.button('Continue', type: 'submit', class: 'button t-continue') %>
       </p>
     <% end %>
+    <p>This service is in development.</p>
+
+    <p>You may not get one of the 3 dates you’ve requested. <br>If there’s a  problem, we’ll call or email to confirm an appointment date with you.</p>
```
benlovell: Should that be a separate paragraph or is the line spacing important?

app/views/booking_requests/step_two.html.erb
```diff
@@ -130,7 +130,7 @@
           <% if @booking_request.errors.include?(:accessibility_requirements) %>
             <span class="error-message"><%= @booking_request.errors[:accessibility_requirements].to_sentence %></span>
           <% end %>
-          <%= f.check_box :accessibility_requirements, class: 't-accessibility-requirements' %> I have accessibility needs
+          <%= f.check_box :accessibility_requirements, class: 't-accessibility-requirements' %> I have accessibility needs <span style="color:#6f777b">eg wheelchair access</span>
```
benlovell: That style thing shouldn't be declared inline but I'll defer to @benbarnett or @Guntrisoft.

benbarnett: Using colour like that isn't part of the GDS style, but you can use the hint-text thing as described here: http://govuk-elements.herokuapp.com/form-elements/#form-hint-text

Guntrisoft: @benbarnett Aman attempted to use a form hint but because it's a block element it was wrapping underneath the checkbox and label - @amansinghb get it fired up on your machine, change it to a form hint and we'll take a look at it together to get it displayed inline.

app/views/booking_requests/step_one.html.erb
amansinghb: @benlovell I showed it to content guys all as one paragraph but after changing it around it a few times it was thought this would be the best for it to be displayed.

@benbarnett I'll sort the extra space out

PR #678, opened 2016-11-05
Capture `date_of_birth` when booking

benlovell:
Makes the `date_of_birth` mandatory. The appointment type is now
inferred from the given date of birth. Both attributes are still passed
to the bookings API to maintain compatibility.

The validation works as you'd expect. When an invalid date (missing a
component) is given this does not trigger the eligibility check. When a
valid (as-in all components of the date are provided) the eligibility
check is triggered, based off the inferred appointment type.

The form looks pretty bad at the moment as I lifted it almost verbatim from
tap. @Guntrisoft - work your magic?


Guntrisoft:
:+1

PR #682, opened 2016-11-11
Content to reinforce select appointments may not be available

amansinghb:
Content to reinforce select appointments may not be available and
consistency around e.g. on the customer details page


benlovell:
For reference: 

screen shot 2016-11-11 at 15 04 25

PR #685, opened 2016-11-16
Bump the quality of static maps

benlovell:
Prior to this change they were a little fuzzy as they were being scaled
as `1` instead of `2` which is our maximum permitted quality on the
free plan.

### Before

screen shot 2016-11-16 at 13 16 04

### After

screen shot 2016-11-16 at 13 15 18

### DIFF
--- a/app/views/locations/index.html.erb
+++ b/app/views/locations/index.html.erb
@@ -38,6 +38,7 @@
           <a href="https://maps.google.com/maps?q=<%= location.address_flattened %>" target="_blank">
             <img
               height="300"
+              alt="map showing the location of <%= location.name %> (link opens external website in new window)"
               style="border:0; position: relative;"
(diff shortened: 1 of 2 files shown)


app/views/locations/index.html.erb
```diff
@@ -39,7 +39,7 @@
             <img
```
benbarnett: I know this isn't really to do with this PR but can we add this alt text to the ``?

`map showing the location of <<Location Name>> (link opens external website in new window)`

Guntrisoft:
This happens on the widest mobile viewport now

screen shot 2016-11-16 at 11 20 16

Guntrisoft:
perhaps a max-width needed

app/views/locations/index.html.erb
benlovell: Cool, will address.

Guntrisoft:
@benlovell will address layout optimisation in future PR

PR #688, opened 2016-11-21
Ensure button is disabled after submission

benlovell:
screen shot 2016-11-21 at 12 49 58


Guntrisoft:
@benbarnett @benlovell Maybe we should write something simple, to disable the button rather than bringing in a whole library

benlovell:
Maybe, although it weighs 20k uncompressed.

Guntrisoft:
Just as an audit trail, I rewrote this branch to use a small bit of JS to disable the button on submit of the form.

PR #703, opened 2016-12-06
Improved accessibility of mobile version of the site navigation

Guntrisoft:
screen shot 2016-12-06 at 13 00 01

- Altered menu link to be a button which better describes
  that it's going to do something interactive when clicked on
- Added aria-expanded=true/false to read out to screen reader
  users that the item that it controls is collapsed, and reads
  out that it's expanded when clicked on
- Added aria-hidden=true on the menu items by default, and then
  turned this to true when button is clicked on, so that menu
  items are only read out when the menu is expanded
- Added aria-controls for screenreaders which support it, so
  a user can jump directly to the menu from the button

### DIFF
--- a/app/assets/javascripts/mobile-nav-toggler.js
+++ b/app/assets/javascripts/mobile-nav-toggler.js
@@ -8,7 +8,16 @@
 
       $trigger.click(function(e) {
         e.preventDefault();
-        $target.toggleClass('active');
+
+        if ($target.hasClass('active')) {
+          $target.removeClass('active');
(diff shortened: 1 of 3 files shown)


benbarnett:
Can explain the reason for this change rather than summarise the changes in the code? We can see the latter from the diff so it's just useful to know what this was for 👍

Guntrisoft:
@benbarnett Sorry, really bad habit of mine 🙈  - fixed.

benbarnett:
Awesome, that's so much clearer now :)

PR #704, opened 2016-12-06
Website amends

Guntrisoft:
screen shot 2016-12-06 at 13 53 19

### DIFF
--- a/app/views/appointment_summaries/new.html.erb
+++ b/app/views/appointment_summaries/new.html.erb
@@ -1,3 +1,4 @@
+<% content_for(:page_title, t('service.title', page_title: 'How to view your appointment summary')) %>
 <% content_for(:meta_description, 'Choose the information you want in the summary of your Pension Wise appointment') %>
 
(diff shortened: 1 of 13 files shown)


app/views/appointment_summaries/new.html.erb
```diff
@@ -1,3 +1,4 @@
+<% content_for(:page_title, 'How to view your appointment summary - Pension Wise') %>
```
benbarnett: Can we move these into the locale file as per the other pages?

Also just noticed that the F2F booking pages are all missing page titles too!

Guntrisoft: I've used the t('service.title', X) pattern for these missing page titles - gone through hopefully all pages that are missing a title and added one.

Guntrisoft:
@alexkouz have a 👀

app/views/booking_requests/step_two.html.erb
```diff
@@ -1,3 +1,4 @@
+<% content_for(:page_title, t('service.title', page_title: "Your details")) %>
```
benlovell: Quotes.

Guntrisoft: Damnation

PR #716, opened 2016-12-29
Fix redirection bug

dwhenry:
When redirecting to a string the following URL is used:

    # actionpack-4.2.7.1 lib/action_controller/metal/redirecting:89
    request.protocol + request.host_with_port + options

This means `redirect_to 'pension-type-tool/question-1'` was
redirecting to:

    https://staging-www.pensionwise.gov.ukpension-type-tool/question-1

Not sure why this issue did not present locally.


benlovell:
Can we add a test to cover this so we don't regress?

PR #722, opened 2017-01-04
Modernise buildpack configuration

benlovell:
The multi buildpack has ceased to be. Heroku now supports multiple
buildpacks as first-class citizens so we should use that style of
configuration instead.


AndrewVos:
LG

PR #725, opened 2017-01-04
New imagery and colour scheme for Facebook landing page

Guntrisoft:
screen shot 2017-01-04 at 16 01 25

- Changes made from feedback from Marketing
- Copy changes after feedback from Content team
- Ensure that transcript content has aria-hidden=true only when
JS is enabled
- Make auto linking of phone number on safari ios white in color


app/views/marketing/facebook-landing.html.erb
```diff
@@ -40,17 +40,17 @@
...
             <h2 class="marketing-subheading">Donna Reid - Pension Wise Guidance Specialist</h2>
 
-            <p>“Pension Wise is a government-backed service that provides free, impartial guidance to help you understand your options for your pensions savings.</p>
+            <p>“Pension Wise is a government-backed service that provides free, impartial guidance to help you understand your options for your pensions savings.”</p>
 
-            <p>“Who is Pension Wise for?</p>
+            <p>Who is Pension Wise for?</p>
 
-            <p>“You're eligible for a free Pension Wise appointment if you're 50 or over, and have defined contribution pension, which is where you build up a 'pot' of savings that you can access when you reach retirement. This would include many private and workplace pensions.</p>
+            <p>“You're eligible for a free Pension Wise appointment if you're 50 or over, and have defined contribution pension, which is where you build up a 'pot' of savings that you can access when you reach retirement. This would include many private and workplace pensions.”</p>
```
AndrewVos: Are you sure you want to be using those quotes in html? Not sure if you need to use special characters?

Guntrisoft: have html escaped them just in case

PR #728, opened 2017-01-04
Whitelist local assets only during JS tests

benlovell:
This stops our tests loading remote assets, speeding up the tests
generally but also silencing the stupid console output from smart
survey's scripts.


AndrewVos:
Nice didn't realise you could (or had to) do this!

Apparently they are fixing it, but this is a sensible default in my opinion.

benlovell:
Yeah, it makes sense for us to default our poltergeist configurations this way across the board. 👍

PR #730, opened 2017-01-05
Add feedback for pension-type-tool

dwhenry:
* Refactor BookingFeedbackForm to FeedbackForm
* Add govspeak renderer


app/forms/feedback_form.rb
```diff
@@ -0,0 +1,31 @@
+class FeedbackForm
+  include ActiveModel::Model
+
+  attr_accessor :name, :email, :message, :booking_type
+
+  validates :name, presence: true
+  validates :email, presence: true, format: { with: /.+@.+\..+/ }
+  validates :message, presence: true
+  validates :booking_type, inclusion: { in: %w(online_booking pension_type_tool) }
```
benbarnett: Would consider renaming `booking_type` to something else – not sure the pension type tool is that kind of thing.

Not quite sure what it should be called though!

app/controllers/booking_requests_controller.rb
```diff
@@ -5,17 +5,17 @@ class BookingRequestsController < ApplicationController
   before_action :set_breadcrumbs
 
   def step_one
-    @booking_feedback = BookingFeedbackForm.new
+    @feedback = FeedbackForm.new(booking_type: 'online_booking')
```
benlovell: Is there any mileage in defaulting this argument to 'online_booking' so we don't have to specify it everywhere? It's overwhelmingly the normal value passed. Or perhaps a specialised method that initialises the object with that `booking_type`. Perhaps `FeedbackForm.for_online_booking` or something?

app/forms/feedback_form.rb
benlovell: Agreed - perhaps `subject`?

dwhenry: sorry that was supposed to be `feedback_type`

benlovell: Makes sense to me.

dwhenry:
@benbarnett @benlovell done and done..

PR #732, opened 2017-01-09
Strip cookies for guide paths

benlovell:
The guides were cacheable prior to 8bfd9f1 that added CSRF protection,
requiring cookies to be enabled. After 8bfd9f1 I noticed a great deal
more traffic at the origin. This will no doubt affect us over the next
few weeks as more traffic arrives due to marketing attempts.

Cloudflare will not cache content returned from the origin when the
`Set-Cookie` response header is present. To restore public caching for
the guide pages we can simply strip away cookies from the request /
response - leaving the other pages to process cookies as normal.

The middleware is loaded early on during the boot process (before Rails
has autoloaded) thus the need to explicitly require the
`GuideRepository` and its dependents.

### DIFF
--- a/app/middleware/strip_session_cookie.rb
+++ b/app/middleware/strip_session_cookie.rb
@@ -1,0 +1,31 @@
+module Middleware
+  class StripSessionCookie
+    def initialize(app, paths:)
+      @app   = app
+      @paths = paths
+    end
+
+    def call(env)
+      if strip?(env)
+        cookieless_call(env)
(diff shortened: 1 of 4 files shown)


benbarnett:
Will this affect anything to do with GTM/GA?

benlovell:
They set cookies for their domain(s), not ours. So we're good. 👍

AndrewVos:
😍

PR #736, opened 2017-01-12
Alter dataLayer push values for player state change

Guntrisoft:
- Feedback from the analytics team suggested we needed to
  push slightly different values

### DIFF
--- a/app/assets/javascripts/video-player.js.es6
+++ b/app/assets/javascripts/video-player.js.es6
@@ -67,9 +67,18 @@
           const originalOnPlayerStateChange = this.onPlayerStateChange;
 
           this.onPlayerStateChange = function(stateId) {
+            let stateValue = 'UNKNOWN';
+
+            $.each(YT.PlayerState, function(key, value) {
+              if (value === stateId) {
+                stateValue = key;
+                return false;
+              }
+            });
+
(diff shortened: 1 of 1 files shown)


benlovell:
We test for the presence of these quite consistently elsewhere. Is that something we should be doing here too?

Guntrisoft:
@benlovell as agreed, testing the change is rather tricky, and perhaps not worth the time to get it working

@benbarnett have added a `return` when a value is matched

app/assets/javascripts/video-player.js.es6
```diff
@@ -67,9 +67,18 @@
           const originalOnPlayerStateChange = this.onPlayerStateChange;
 
           this.onPlayerStateChange = function(stateId) {
+            let stateValue = 'UNKNOWN';
+
+            $.each(YT.PlayerState, function(key, value) {
+              if (value === stateId) {
+                stateValue = key;
+                return;
```
benbarnett: Tiny thing but I think we need to actually `return false;` out of this: http://stackoverflow.com/questions/1799284/how-to-break-exit-from-a-each-function-in-jquery

Guntrisoft: sorted

PR #738, opened 2017-01-13
Link to the three online booking regions for Tesco pilot

benbarnett:


content/tesco.md
```diff
@@ -32,4 +32,8 @@ Afterwards you’ll get a summary of what was discussed.
 
 ^Pension Wise guidance specialists are impartial. They don’t recommend any products or companies and won’t tell you how to invest your money.^
 
-You can book your [Pension Wise appointment online](https://tpw.acuityscheduling.com/schedule.php#) now.
+To book your Pension Wise appointment online, use the link below to select your location:
+
+[England & Wales](https://tpw.acuityscheduling.com)
```
benlovell: Both this link and the bottom one actually 302 to `schedule.php` so perhaps we should just add that to the link and avoid an unnecessary roundtrip?

benbarnett: Ah, nice catch will do

benlovell:
Squishy merge 🚢

benbarnett:
Cheers. Squashed down, just waiting for all clear from the team running the pilot.

PR #740, opened 2017-01-16
Use "dogs" campaign artwork on homepage

benbarnett:
This replaces the previous "coffee cups" artwork, as the campaign is no longer running.


benlovell:
The code is a bit ruff in places. 🐶

PR #741, opened 2017-01-18
Fix issue with screen readers not able to see navigation

Guntrisoft:
- Desktop screen reader users would not have been able to
  access the site navigation as it had a aria-hidden=true
  on it
- Rewritten navigation JS as es6 module
- Module now makes nav aria-hidden only on mobile, until
  they click on the menu button


app/assets/javascripts/components/Navigation.es6
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    setNavigationHiddenStatus() {
+      this.$target.attr('aria-hidden', this.getPageSize() === this.mobileSize ? true : false);
+    }
+
+    bindEvents() {
+      $(window).on('resize', this.debounce(this.setNavigationHiddenStatus.bind(this), 50));
+
+      this.$trigger.click(this.handleClick.bind(this));
```
benbarnett: Can we use the `on` style here too for consistency?

app/assets/javascripts/components/Navigation.es6
```diff
@@ -0,0 +1,53 @@
...
+    constructor($component) {
+      super($component);
+      this.init();
+    }
+
+    init() {
+      this.$trigger = $('.js-nav-toggler');
+      this.$target = $('.js-nav');
+      this.$nav = $('.l-page-nav');
+      this.mobileSize = '"mobile"';
```
benbarnett: 🤢

app/assets/javascripts/components/Navigation.es6
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    getPageSize() {
+      const el = window.getComputedStyle(this.$nav[0], ':after');
+
+      return el ? el.getPropertyValue('content') : this.mobileSize;
+    }
+
+    setNavigationHiddenStatus() {
+      this.$target.attr('aria-hidden', this.getPageSize() === this.mobileSize ? true : false);
```
benbarnett: You could use `this.getPageSize().indexOf(this.mobileSize)` here to get around that double quote shenanigan

app/assets/javascripts/components/Navigation.es6
Guntrisoft: Yup, agree it's nasty!

app/assets/javascripts/components/Navigation.es6
Guntrisoft: Done

app/assets/javascripts/components/Navigation.es6
Guntrisoft: Done

Guntrisoft:
@benbarnett added aria tests

app/assets/javascripts/components/BaseComponent.es6
```diff
@@ -11,6 +11,26 @@
...
+            timeout = null;
+            if (!immediate) func.apply(context, args);
+          },
+          callNow = immediate && !timeout;
+
+        clearTimeout(timeout);
+
+        timeout = setTimeout(later, wait);
+
+        if (callNow) func.apply(context, args);
```
benbarnett: Actually thought this wouldn't get through the linting. What do you reckon to enabling the [curly](http://eslint.org/docs/rules/curly) rule? Any opinions either way?

Guntrisoft: yup, that seems fair enough - will sort

Guntrisoft: Done

benbarnett:
Cool, but we should separate out the linting config change into its own commit, apply that first, and then add the code change.

We should also explain why we're making the change, even if it's just to explain that we're linting for a standard style the team has already adopted.

Guntrisoft:
@benbarnett agreed - have rebased to split it out

PR #742, opened 2017-01-19
Validate customer emails during bookings process

benlovell:
This adds some guarantees that the format of customer-provided emails
are at least some way correct.


AndrewVos:
Looks good

PR #743, opened 2017-01-19
Resolve email validation issues

benlovell:
Ensures the correct validation message is displayed when the email is
provided incorrectly.

Also removes the redundant and unused validator that clashed with the gem's version.


Guntrisoft:
👍

PR #748, opened 2017-01-26
Allow feedback to be left on the pension type tool

dwhenry:
This was failing to work as the cookies are striped
from the page to enable caching. This change requires
CSRF protection from the feedback submission form.

### DIFF
--- a/spec/features/feedback_spec.rb
+++ b/spec/features/feedback_spec.rb
@@ -1,0 +1,24 @@
+require 'spec_helper'
+
+RSpec.feature 'Customer feedback', js: true do
+  scenario 'can be left on the pension type tool' do
+    begin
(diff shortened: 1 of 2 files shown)


spec/features/feedback_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+RSpec.feature 'Customer feedback', js: true do
+  around do |example|
```
benlovell: This isn't actually doing anything as you need to call `example.run`. Also may as well just inline it into the single `scenario` where it's used.

dwhenry: Will inline.. but just so you know.. `run` is an alias of `call` :) https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/example.rb#L340

PR #750, opened 2017-01-27
Alternative landing page available at `/landing`

benbarnett:
The page is identical to the existing 'facebook landing' page,
except this one is to be used for various other digital marketing
and no phone number to promote online booking.


Guntrisoft:
@benbarnett just noticed that there is a redundant action in the marketing controller which probably could be tidied up as part of this PR?

benbarnett:
@Guntrisoft might just do that separately as there still seems to be some discussion around this page.

benbarnett:
Addressed in https://github.com/guidance-guarantee-programme/pension_guidance/pull/755

PR #753, opened 2017-01-30
Namespacing IDs for form elements

Guntrisoft:
- avoid duplication of element IDs
- makes page HTML valid

### DIFF
--- a/app/views/appointment_summaries/create.html.erb
+++ b/app/views/appointment_summaries/create.html.erb
@@ -6,7 +6,7 @@
 
 <div class="fixed-panel js-fixed-panel hide-from-print">
   <div class="fixed-panel__button">
-    <%= form_for @appointment_summary, url: download_appointment_summaries_path do |f| %>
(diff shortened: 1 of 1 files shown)


app/views/appointment_summaries/create.html.erb
```diff
@@ -17,7 +17,7 @@
   </div>
 
   <div class="fixed-panel__button">
-    <%= form_for @appointment_summary, url: print_appointment_summaries_path, html: { target: "_blank" } do |f| %>
+    <%= form_for @appointment_summary, url: print_appointment_summaries_path, html: { target: "_blank", id: 'print_form' }, namespace: "print" do |f| %>
```
benbarnett: Got a bit of double / single quotes going on here

benbarnett:
Much nicer commit now ;) Just that comment about the " and 's and 👍

app/views/appointment_summaries/create.html.erb
Guntrisoft: Done!

PR #755, opened 2017-01-31
Remove redundant action in marketing controller

benbarnett:
Was accidentally left over from a previous iteration.


Guntrisoft:
👍 💯

PR #756, opened 2017-01-31
Adding in aXe accessibility testing

Guntrisoft:
axelogo

Using axe-matchers gem to integrate accessibility tests into a set of new cucumber features.

The tests which are run against the pages are listed here:
https://github.com/dequelabs/axe-core/blob/master/doc/rule-descriptions.md

Altered steps which select radio buttons on appointment summary form page to ensure they work with features with/without Javascript

### DIFF
--- a/features/accessibility.feature
+++ b/features/accessibility.feature
@@ -1,0 +1,33 @@
+@javascript
+Feature: Accessibility of pages
+  As a customer I want Pension Wise pages to be fully accessible
+  so I can navigate the website regardless of my access needs
+
+  Scenario: Homepage of the site is accessible
+    When I visit the homepage
+    Then the page should be accessible
+
+  Scenario Outline: Typical pages of the site are accessible
+    When I visit the page <Slug>
+    Then the page should be accessible
+
+    Examples:
+      | Slug                     |
+      | facebook-landing         |
+      | leave-pot-untouched      |
+      | tax                      |
+      | summary-document/new     |
+      | feedback/new             |
+
+  Scenario: Location page is accessible
+    When I visit a face to face booking location page
+    Then the page should be accessible
+
+  Scenario: Appointment summary page is accessible
+    When I visit the appointment summary page
+    Then the page should be accessible
+
+  @booking_locations
+  Scenario: Slot picker for face to face booking is not accessible while we develop our own slot picker
+    When I view the face to face booking form
+    Then the page should be accessible excluding ".SlotPicker"
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ group :development do
 end
 
 group :test do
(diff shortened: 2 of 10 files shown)


features/accessibility.feature
```diff
@@ -0,0 +1,39 @@
+@javascript
+Feature: Accessibility of pages
+  As a customer I want Pension Wise pages to be fully accessible
+  so I can navigate the website regardless of my disability
```
benbarnett: `regardless of my access needs` maybe reads a little more clearly?

features/accessibility.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  @vcr
+  Scenario: Location page is accessible
+    Given I have searched for appointment locations near to a valid postcode
+    When I drill down into a specific search result
+    Then the page should be accessible
+
+  Scenario: Appointment summary page is not accessible until errors have been fixed
+    When I visit the digital appointment summary generator
+    And I select 55 or over as my age range
```
benbarnett: We could do `And I select any age range` and add a step that'll do a `.sample` between the two options maybe?

benbarnett: Up for discussion, not sure how much value this'll give in the scheme of things

features/accessibility.feature
```diff
@@ -0,0 +1,39 @@
...
+    And I view the appointment summary
+    Then the page should be accessible
+
+  @booking_locations
+  Scenario: Slot picker for face to face booking is not accessible while we develop our own slot picker
+    Given a location is enabled for online booking
+    When I browse for the location
+    And I opt to book online
+    Then the page should not be accessible
```
benbarnett: Was there another way of wording this that better described our situation? I think it was just an alias for this one.

benbarnett:
Loving this so far

features/accessibility.feature
Guntrisoft: Done

features/accessibility.feature
Guntrisoft: I'm kinda questioning the actual feature specs to be honest - as i'm piecing together some steps in order to get to the page I want to test for accessibility.

Maybe I need something a lot simpler as a feature spec, which then follows some steps in the background.

features/accessibility.feature
Guntrisoft: I'm struggling to find what we saw yesterday in the docs!

Guntrisoft: I've altered so it excludes the slot picker but tests the rest of the page

features/accessibility.feature
Guntrisoft: @benbarnett I've decided to simplify the feature specs, and keep the details of how I get to the particular page to the step definition - let me know what you think

PR #757, opened 2017-02-01
Pass 'not sure' through to bookings API

benlovell:
To facilitate displaying when the customer specified they were unsure
whether their pension was DC, we now pass their selection through to
the bookings API per the requirements for displaying in planner.


Guntrisoft:
👍

PR #759, opened 2017-02-07
Show error message when customer does not enter a postcode

Guntrisoft:
screen shot 2017-02-07 at 12 18 09

If a user does not enter a postcode on face to face location
search then we show the location search again but include
a suitable error message to inform the user what they need
to do

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -8,7 +8,8 @@ class LocationsController < ApplicationController
   layout 'full_width_with_breadcrumbs', only: %i(show index)
 
   def index
-    return render :search unless @postcode.present?
+    return render :search if @postcode.nil?
+    return render :empty_postcode if @postcode.empty?
 
     @locations = begin
       Locations.nearest_to_postcode(@postcode, limit: NEAREST_LIMIT).map do |location|
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -11,15 +11,7 @@ class LocationsController < ApplicationController
     return render :search if @postcode.nil?
     return render :empty_postcode if @postcode.empty?
 
-    @locations = begin
-      Locations.nearest_to_postcode(@postcode, limit: NEAREST_LIMIT).map do |location|
-        LocationSearchResultDecorator.new(location)
-      end
-    rescue Geocoder::InvalidPostcode
-      render :invalid_postcode
-    rescue Geocoder::FailedLookup
-      render :failed_lookup
-    end
+    retrieve_locations
   end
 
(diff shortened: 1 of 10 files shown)


Guntrisoft:
@alexkouz is the wording ok here in the error message (see screenshot above)

app/views/locations/empty_postcode.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render 'components/error_postcode' do %>
+  <span class="t-empty-postcode error-message">Please enter a valid postcode</span>
+<% end %>
+
```
benlovell: Nope.

features/find_nearest_face-to-face_appointment_locations.feature
```diff
@@ -12,6 +12,10 @@ Scenario: Search using an invalid postcode (i.e. one that we can't find)
   When I search for appointment locations near to an invalid postcode
   Then I should be informed that Pension Wise cannot find that postcode
 
+Scenario: Search without entering a postcode
+  When I search for appointment locations without entering a postcode
+  Then I should be informed that I need to enter a postcode
```
benlovell: 'Then I am told to enter a valid postcode' would be better I think.

app/controllers/locations_controller.rb
```diff
@@ -8,7 +8,8 @@ class LocationsController < ApplicationController
   layout 'full_width_with_breadcrumbs', only: %i(show index)
 
   def index
-    return render :search unless @postcode.present?
+    return render :search if @postcode.nil?
+    return render :empty_postcode if @postcode.empty?
```
benlovell: Maybe we could put a client validation pattern or something on the frontend to stop this roundtrip for a large number of cases too?

features/find_nearest_face-to-face_appointment_locations.feature
Guntrisoft: Done

app/views/locations/empty_postcode.html.erb
Guntrisoft: Yup.

app/controllers/locations_controller.rb
Guntrisoft: Added `required=true`

Guntrisoft:
@benlovell https://github.com/guidance-guarantee-programme/pension_guidance/pull/759/files#diff-54b4e4f2b562d74dad3af78e66e74e4fR14 - this ok? (to avoid rubocop issue of too many lines for `def index`

alexkouz:
@Guntrisoft yep, happy with the above postcode error message

PR #763, opened 2017-02-08
Ensure a DC eligibility option is chosen

benlovell:
There was no validation being performed on this form value prior to
this change.

screen shot 2017-02-08 at 15 04 45

screen shot 2017-02-08 at 15 04 52


Guntrisoft:
`can't be blank` seems like odd terminology to me

benlovell:
@Guntrisoft That's standard Rails terminology, used everywhere else when a value is not provided. Do you have a better suggestion?

Guntrisoft:
Agreed with content team this is ok.

Guntrisoft:
👍

PR #767, opened 2017-02-13
Update copy on telephone booking confirmation page

Guntrisoft:
User research found that users were finding the page
ambiguous in terms of what they would receive in the
confirmation email and they would like to know the
details of their appointment in case the email
didn't reach them.

Switched to two thirds layout which is the same as
face to face booking request confirmation page

screen shot 2017-02-13 at 09 33 53

### DIFF
--- a/app/views/telephone_appointments/confirmation.html.erb
+++ b/app/views/telephone_appointments/confirmation.html.erb
@@ -1,17 +1,19 @@
-<h1>Your booking is confirmed</h1>
+<div class="l-grid-row">
+  <div class="l-column-two-thirds">
+    <h1>We’ve booked your appointment</h1>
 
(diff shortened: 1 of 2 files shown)


benlovell:
Why did we do this?

Guntrisoft:
@alexkouz Please confirm you are happy with the above

Guntrisoft:
@CassBonner has confirmed this is ok

Guntrisoft:
@benlovell sorry, updated with more info

benbarnett:
Can we add something about "in user research we found that users were finding the page ambiguous in terms of what they would receive in the email, and that they would like to know the details of their appointment in case it doesn't come through" or something

PR #768, opened 2017-02-13
Update Annuity calculator constants - Feb 2017

Guntrisoft:
Update annuity rates with latest data


benbarnett:
Wow, didn't expect those to generate the same numbers. Nice one.

PR #771, opened 2017-02-20
Warn when a question is unanswered

benlovell:
When the customer forgets to supply an answer on any of the pension type
tool questions we currently just redirect back, guarding them from
proceeding but without any explanatory text.

This change adds a flash warning message above the heading asking the
customer to complete and submit their answer correctly.

screen shot 2017-02-20 at 15 10 26

### DIFF
--- a/app/repositories/guide_repository.rb
+++ b/app/repositories/guide_repository.rb
@@ -11,6 +11,10 @@ class GuideRepository
     @slugs ||= new.slugs
   end
 
+  def self.cacheable_slugs
+    slugs.reject { |s| /question-\d+\Z/ === s } # rubocop:disable Style/CaseEquality
+  end
+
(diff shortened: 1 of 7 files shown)


app/repositories/guide_repository.rb
```diff
@@ -11,6 +11,10 @@ def self.slugs
     @slugs ||= new.slugs
   end
 
+  def self.cacheable_slugs
```
benlovell: We needed to strip the question pages here so cookies are stripped and flash works.

Guntrisoft:
@benlovell I answered the first question, and on the `/pension-type-tool/question-2` page I was presented with the flash message

screen shot 2017-02-20 at 14 03 49

Guntrisoft:
The issue surrounding my comment above has now been fixed

LGTM 👍

PR #778, opened 2017-02-27
Bump outdated gems

benlovell:
Bumps gem dependencies via `bundle outdated`. We should do this
regularly to stay on top of bug fixes and security releases, and to
ease future upgrades.


Guntrisoft:
Nice - had it on my list to update frontendtoolkit today 👍

PR #780, opened 2017-02-28
Accommodate per-location slots for face-to-face

benlovell:
We need to display per-location slots for face-to-face bookings due to
the upcoming location closures. These changes ensure the calendar is
bound to the chosen location's slots, not its booking location as
before.


benlovell:
[As a customer I am not able to request an appointment for after the 31st of March at certain locations](https://trello.com/c/2X5DbcL8/18-as-a-customer-i-am-not-able-to-request-an-appointment-for-after-the-31st-of-march-at-certain-locations)

PR #788, opened 2017-03-03
Cache booking locations with redis store

benlovell:
Using the redis store instead of the memory store means the cache
survives app restarts and is shared across however many web processes
we are using. This is favourable to each web process maintaining its
own in-memory cache.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,4 +73,5 @@ end
 
 group :staging, :production do
   gem 'rails_12factor'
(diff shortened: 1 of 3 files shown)


benlovell:
[Locations cache becomes corrupted in background workers](https://trello.com/c/MP00fdh3/38-locations-cache-becomes-corrupted-in-background-workers)

Guntrisoft:
👍

PR #793, opened 2017-03-08
Bump `booking_locations` for cache expiry

benlovell:
Ensure the cache is evicted with respect to the configured TTL.


benlovell:
[Ensure TTL is respected on cached booking location responses](https://trello.com/c/ApuBJ5Ak/47-ensure-ttl-is-respected-on-cached-booking-location-responses)

PR #795, opened 2017-03-09
Alter headers in footer of site to H2s / Guider sidebar to <aside>

Guntrisoft:
The headers in the footer are better placed to be `H2`s as a descendant of the main header of the page (`H1`).

As `H3`s they ended up being a sub header of a random section of content which they were completely unrelated to.

Alter guide sidebar to be an `<aside>` element to aid screen readers.

Removes H2 header from the guider page right column as it is a repeat of the H1 of the page, and adds confusion.

**Example Document outline after change**
screen shot 2017-03-09 at 13 53 59

### DIFF
--- a/app/assets/stylesheets/components/_footer-categories.scss
+++ b/app/assets/stylesheets/components/_footer-categories.scss
@@ -12,6 +12,14 @@
       margin: 0 30px;
     }
   }
+
+  h2 {
+    @include bold-19;
+
+    @include media(tablet) {
+      margin: 0 15px;
+    }
+  }
 }
 
 .footer-categories {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/components/_footer-categories.scss
```diff
@@ -12,6 +12,16 @@
       margin: 0 30px;
     }
   }
+
+  h2 {
+    font-size: 16px;
+
+    @include media(tablet) {
+      font-size: 19px;
```
benbarnett: is there a mixin for this stuff somewhere?

Guntrisoft: Yup - you're right, there is. added.

PR #797, opened 2017-03-10
Convert step circles css code to not use images

Guntrisoft:
There was a lot of css produced to output the step
icons, these could be recreated in pure css with css counters.

As this has been done in pure css then there is no need for
all the step images in the project.

**Chrome on Mac**
screen shot 2017-03-10 at 13 24 46

**IE8 and less fallback**
screen shot 2017-03-10 at 13 45 40


jongilbraith:
Nice 👍

PR #799, opened 2017-03-10
Bump Ruby to 2.4

benlovell:
Take advantage of incremental improvements to the Ruby GC and language.

I had to pin `rubytree` at our fork while a PR is pending for 2.4
related fixes.

Also used `Regex#match?` as it was recommended by Rubocop's 2.4 syntax
check.


jongilbraith:
👍

PR #801, opened 2017-03-12
Survive refreshes using query string parameters

benlovell:
The confirmation page was relying on flash parameters being set by the
referring action, when customers refreshed the page their session would
generally be expired and an error would occur.

Passing these values along with the query string works around the issue.


Guntrisoft:
:+1:

PR #802, opened 2017-03-13
Differentiate F2F and telephone booking feedback

benlovell:
We are using separate surveys to track feedback for both channels.


benlovell:
[As a customer who has booked a telephone appointment online I want to give feedback on the booking journey](https://trello.com/c/0hhyRaKm/50-as-a-customer-who-has-booked-a-telephone-appointment-online-i-want-to-give-feedback-on-the-booking-journey)

PR #803, opened 2017-03-13
Fix a timing issue in the booking feature

benlovell:
We need to explicitly wait for the slots to be bound here before
asserting on their presence.


jongilbraith:
👍

PR #807, opened 2017-03-16
Check F2F fields are valid before ineligible

benlovell:
Prior to this change we were overzealously checking eligibility
regardless whether the fields needed to determine these facts were
supplied by the customer.

This simple change just checks the values are valid before applying the
eligibility rules.


benlovell:
[As a customer booking an online appointment I am shown an error message if I forget to fill in any of the details on the booking form](https://trello.com/c/JSvLk3dh/63-as-a-customer-booking-an-online-appointment-i-am-shown-an-error-message-if-i-forget-to-fill-in-any-of-the-details-on-the-booking)

PR #808, opened 2017-03-17
Expose an API for clearing booking locations

benlovell:
We need a way of remotely triggering a cache flush when booking
locations change structurely. This change adds an API endpoint, with
simple token based authentication. We are not using SSO here since this
application is not backed by a persistent store - a requirement for
GDS-SSO and signon integration.

A request issued to `DELETE /locations_cache` with the correct
authorisation token will sweep the cached locations and leave
non-prefixed (anything other than our locations) entries as-is.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -82,7 +82,7 @@ GEM
     babel-transpiler (0.7.0)
       babel-source (>= 4.0, < 6)
       execjs (~> 2.0)
-    booking_locations (0.9.1)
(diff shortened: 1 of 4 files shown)


benlovell:
[As an admin I am able to move a child location under one parent to another](https://trello.com/c/WHspsUZV/43-as-an-admin-i-am-able-to-move-a-child-location-under-one-parent-to-another)

PR #826, opened 2017-03-27
Add book online button to phone page

Guntrisoft:
After a successful trial of adding the online booking button to the `book-phone` page, this will add the button permanently to the page - thus officially launching online telephone booking 🎉 

screen shot 2017-03-27 at 09 39 26

### DIFF
--- a/content/book_phone.md
+++ b/content/book_phone.md
@@ -4,12 +4,13 @@ tags:
   - appointments
   - booking
 ---
-
 # How to book a phone appointment
 
-Phone **0800 138 3944**.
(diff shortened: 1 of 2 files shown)


content/book_phone.md
```diff
@@ -4,12 +4,13 @@ tags:
   - appointments
   - booking
 ---
-
 # How to book a phone appointment
 
-Phone **0800 138 3944**.
+[Book online](/telephone_appointments/new){: .button #phone-button style="font-size: 0.9em; padding-left: 2em; padding-right: 2em;"}
```
benlovell: Can you make this `/telephone-appointments/new` so we avoid a redirect? Also in a separate commit could you remove the redirect from the routes?

Guntrisoft: Good spot - done!

PR #846, opened 2017-03-27
Precompile assets in CI

benlovell:
This should mitigate those issues where poltergeist times out on the
first 'JS' request due to JIT asset compilation.


benlovell:
This is going to need a heap of work in `jasmine-rails` of which I do not currently have the time :(

PR #847, opened 2017-03-27
Ensure content is cached with a TTL

benlovell:
Up to now content items were being cached indefinitely. This was not a
problem when we were configured to use the memory store that wouldn't
survive application restarts but is now we use an out-of-process cache
store via Redis. This change ensures we respect the TTL on content
items.


benlovell:
[Guides and content should be expired according to the configured TTL](https://trello.com/c/y3DEmtgL/72-guides-and-content-should-be-expired-according-to-the-configured-ttl)

PR #848, opened 2017-03-29
Solve race between calendar and slot bookings

benlovell:
This solves the race condition between a calendar being displayed and
slots being booked for previously available days. In this case we
simply ask the customer to choose another day. When the page posts back
the calendar is reloaded and their originally chosen day will be also
made unavailable.

screen shot 2017-03-29 at 14 40 32


benlovell:
[Guard against slot days and availability disappearing during a customer's telephone booking journey](https://trello.com/c/2ybNhO4Q/79-guard-against-slot-days-and-availability-disappearing-during-a-customer-s-telephone-booking-journey)

Guntrisoft:
💯 👍

PR #862, opened 2017-04-06
Call centre opening time changed

amansinghb:
As part of the new agreement the opening times for the call centre to
book Pension Wise appointment is now 8am - 8pm Monday to Friday.

The new times will need to be reflected on the website where shown.


benlovell:
[As a customer wanting to booking an appointment over the phone I am able to see the new opening times for the call centre](https://trello.com/c/hpsg0gVf/106-as-a-customer-wanting-to-booking-an-appointment-over-the-phone-i-am-able-to-see-the-new-opening-times-for-the-call-centre)

PR #866, opened 2017-04-12
Update customer_promise.md

alexkouz:
Removed reference to being able to get an appointment within 14 days as no longer guaranteed.


Guntrisoft:
👍

PR #867, opened 2017-04-12
Adding a continue button to the telephone appointment form

Guntrisoft:
After a user selected a time we were moving them on automatically
to the customer form page.

Feedback from DWP content team suggested that the standard
gov.uk way is to let a user make a choice and then decide to
continue to the next step by clicking a 'continue' button

screen shot 2017-04-12 at 13 00 06

### DIFF
--- a/app/controllers/telephone_appointments_controller.rb
+++ b/app/controllers/telephone_appointments_controller.rb
@@ -8,6 +8,8 @@ class TelephoneAppointmentsController < ApplicationController
     retrieve_slots
   end
 
+  helper_method :slot_selected?
+
   def new
   end
 
@@ -33,8 +35,10 @@ class TelephoneAppointmentsController < ApplicationController
   def create_step_2
     if request.xhr?
       render partial: 'times', locals: { times: @times }
-    else
+    elsif slot_selected?
       telephone_appointment.advance! { render :new }
+    else
+      render :new
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/controllers/telephone_appointments_controller.rb
```diff
@@ -34,7 +34,11 @@ def create_step_2
     if request.xhr?
       render partial: 'times', locals: { times: @times }
     else
-      telephone_appointment.advance! { render :new }
+      if telephone_appointment.start_at.nil?
```
benlovell: Can we flip this logic around and say `if telephone_appointment.start_at?` instead?

app/views/telephone_appointments/new.html.erb
```diff
@@ -1,18 +1,22 @@
...
           Unable to submit the form
         </h2>
+
         <p>Check the following:</p>
 
         <ul class="error-summary-list">
           <% @telephone_appointment.errors.each do |key, message| %>
             <li><%= link_to "#{TelephoneAppointment.human_attribute_name key} – #{message}", "#telephone_appointment_#{key}" %></li>
           <% end %>
+          <% if @telephone_appointment.start_at.nil? %>
```
benlovell: I think it's clearer to switch this around and state `unless @telephone_appointment.start_at?` here.

app/controllers/telephone_appointments_controller.rb
Guntrisoft: Done

app/views/telephone_appointments/new.html.erb
Guntrisoft: Done

Guntrisoft:
[Add green continue button under times on 'Book a phone appointment' page](https://trello.com/c/4i3IfOQ3/100-add-green-continue-button-under-times-on-book-a-phone-appointment-page)

Guntrisoft:
@benlovell added `slot_selected?` method

PR #870, opened 2017-04-13
Updating calculator error msgs to be inline with other error msgs

Guntrisoft:
This is to ensure that we have a consistent set of error messages
across the site

screen shot 2017-04-13 at 17 36 45


Guntrisoft:
[Make error messages across the rest of site consistent with booking path errors.](https://trello.com/c/l4StC71W/114-make-error-messages-across-the-rest-of-site-consistent-with-booking-path-errors)

PR #872, opened 2017-04-18
Amend error messages for guaranteed income calculator

Guntrisoft:
After feedback from the content team it has been decided to
amend the error message for users who enter an age younger
than 55 or older then 75.

## Older than 75
screen shot 2017-04-18 at 16 02 29

## Younger than 55
screen shot 2017-04-18 at 16 02 34


Guntrisoft:
[Make error messages across the rest of site consistent with booking path errors.](https://trello.com/c/l4StC71W/114-make-error-messages-across-the-rest-of-site-consistent-with-booking-path-errors)

PR #899, opened 2017-05-05
Validate structure of phone numbers

benlovell:
This adds very primitive validation just to ensure we get something
approaching a correct telephone number.


benlovell:
[Validate phone numbers upon entry](https://trello.com/c/eicTOVc6/153-validate-phone-numbers-upon-entry)

PR #900, opened 2017-05-05
Provide friendly validation messages

benlovell:
Without these the advice is harshly worded.


benlovell:
[Validate phone numbers upon entry](https://trello.com/c/eicTOVc6/153-validate-phone-numbers-upon-entry)

PR #924, opened 2017-05-16
Remove google experiments code which prevents page flicker

Guntrisoft:
Now that we have ended the A/B test for the homepage we can
now remove this 'page hiding script'

Originally added in:
https://github.com/guidance-guarantee-programme/pension_guidance/commit/3b8c633b12194fc6184082b8841cf68484d53045


Guntrisoft:
[Turn off page hiding script for A/B - when A/B finishes](https://trello.com/c/Xgd4G321/132-turn-off-page-hiding-script-for-a-b-when-a-b-finishes)

PR #925, opened 2017-05-16
Update homepage to new layout/content

Guntrisoft:
After A/B testing a new layout for the homepage, it has
proved succesful enough to make it the default homepage.

This removes traces of the old homepage and makes
the 'experiment' version the new one.

screen shot 2017-05-16 at 15 33 30


Guntrisoft:
[As a product owner I want to stop the A/B homepage test and set the test winner as the new homepage](https://trello.com/c/avM26ZUd/171-as-a-product-owner-i-want-to-stop-the-a-b-homepage-test-and-set-the-test-winner-as-the-new-homepage)

PR #926, opened 2017-05-17
Update homepage cap style

alexkouz:
Updated bullet sentence style so that when a colon introduces a list, the first word after the colon isn't capitalised.


benlovell:
🎩 👒 style.

PR #930, opened 2017-05-17
Missing translations

jongilbraith:
* Removed what appears to be an unused partial in `components/pager`
* Translated the previous/next buttons in the actually used pagination in `guides/journey_links`
* Translated the loading message in calculators. Slightly crude technique, but seems the best way forward considering it appears to be the only copy produced via js for now.
* Added missing time formatting translations which don't reveal themselves in production, but do when in development.


config/locales/cy.yml
```diff
@@ -57,6 +61,8 @@ cy:
       govuk_date: '%-e %B %Y'
 
   time:
+    am: am
```
benlovell: Can we add a quick spec that attempts to format using these so this doesn't regress?

app/assets/javascripts/calculators.js
```diff
@@ -66,8 +66,10 @@
         this.$loadingStatus.remove();
       }
 
-      this.$loadingStatus = $('<span class="calculator__loading-status">Please wait...</span>').
-                           insertAfter(this.$submitButton);
+      this.$loadingStatus = $('<span class="calculator__loading-status">' +
```
Guntrisoft: 👍 agreed

jongilbraith:
@benlovell comments addressed

PR #932, opened 2017-05-22
Update SP rules

CassLBonner:
Changed lines around SP top up to bring them up to date.


benlovell:
@arwelowen Could this please be translated?

PR #936, opened 2017-05-22
Reset the current locale to `en` after each test

jongilbraith:
Locale state was leaking between specs and causing failures depending
on order. To address this we can simply 'reset' the locale back to
English after each test run in both RSpec and Cucumber.


benlovell:
Can we add a little reasoning on the why you had to do this?

PR #941, opened 2017-06-01
Alter booking form step one to show no availibility message

Guntrisoft:
If a location has no availability then this shows a message
for the customer to call the location instead of showing
an empty calendar.

screen shot 2017-06-01 at 13 41 00


Guntrisoft:
[As a customer wanting to booking an appointment I am informed there is no availability if there is a location with no availability](https://trello.com/c/wzGWNVCH/184-as-a-customer-wanting-to-booking-an-appointment-i-am-informed-there-is-no-availability-if-there-is-a-location-with-no-availabili)

PR #943, opened 2017-06-02
Adding an additional info field to F2F booking form

Guntrisoft:
**This PR needs to be merge after this one https://github.com/guidance-guarantee-programme/planner/pull/169**

This introduces an additional information text field which allows customers to specify an additional amount of information about their booking request e.g. times they aren't available during
a AM/PM slot.

It also introduces a character limit module in JS which

a) Limits input to the textbox to 160 chars on older browsers which don't support the maxlength attribute on a <textarea>
b) Displays a character remaining count underneath the textarea
c) Makes use of the aria-live and aria-atomic attributes when the user is getting low on characters left, so screen reader users are aware that they are starting to hit the character limit.

### DIFF
--- a/app/forms/booking_request_form.rb
+++ b/app/forms/booking_request_form.rb
@@ -4,7 +4,7 @@ class BookingRequestForm
   attr_accessor :location_id, :primary_slot, :secondary_slot, :tertiary_slot,
                 :first_name, :last_name, :email, :telephone_number,
                 :memorable_word, :accessibility_requirements,
-                :date_of_birth, :dc_pot
+                :date_of_birth, :dc_pot, :additional_info
 
   with_options if: :step_one? do |step_one|
     step_one.validates :primary_slot, presence: true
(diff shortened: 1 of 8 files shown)


Guntrisoft:
[As a customer booking a face-to-face appointment I am able to provide additional information that I feel may help when booking](https://trello.com/c/335eC4WV/186-as-a-customer-booking-a-face-to-face-appointment-i-am-able-to-provide-additional-information-that-i-feel-may-help-when-booking)

app/forms/booking_request_form.rb
```diff
@@ -4,7 +4,7 @@ class BookingRequestForm
   attr_accessor :location_id, :primary_slot, :secondary_slot, :tertiary_slot,
                 :first_name, :last_name, :email, :telephone_number,
                 :memorable_word, :accessibility_requirements,
-                :date_of_birth, :dc_pot
+                :date_of_birth, :dc_pot, :additional_info
```
benlovell: We can validate this for length, when provided.

PR #948, opened 2017-06-12
Removing reference to Citizens Advice from all location pages

amansinghb:
As all location addresses have been updated fully we no longer require
the location title and Citizens Advice reference as this does not apply
to a number of locations. The 1st line of the address contains the
information whether its a Age Uk or Citizens advice location.

### DIFF
--- a/app/views/locations/show.html.erb
+++ b/app/views/locations/show.html.erb
@@ -25,7 +25,7 @@
 <div class="l-grid-row">
   <div class="l-column-half">
     <p>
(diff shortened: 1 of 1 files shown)


amansinghb:
[As a customer I want to see the full address of an appointment location so that I know where to go](https://trello.com/c/VkjVxFPT/143-as-a-customer-i-want-to-see-the-full-address-of-an-appointment-location-so-that-i-know-where-to-go)

PR #946, opened 2017-06-08
Limited availability

jongilbraith:
**Limited availability warning on search results**

screen shot 2017-06-08 at 15 04 12

**Alternate location suggestions on booking page**

Locations within 5 miles with 3 or more available slots.

screen shot 2017-06-08 at 15 10 54

Left the two refactoring commits separate for review, will rebase in before merging.

PR #952, opened 2017-06-21
Hide 'Book online' button when a location has no availability

jongilbraith:
Some ground work needed before making the very simple addition of view logic to perform the task at hand, but handily was able to pull over a few of the commits from my parked [limited availability PR](https://github.com/guidance-guarantee-programme/pension_guidance/pull/946).

### DIFF
--- a/spec/lib/booking_requests/booking_requests_spec.rb
+++ b/spec/lib/booking_requests/booking_requests_spec.rb
@@ -2,8 +2,11 @@ RSpec.describe BookingRequests do
   let(:api) { instance_double(BookingRequests::Api) }
 
   describe '.slots' do
+    around do |example|
+      with_stubbed_booking_requests_api(api, example)
+    end
+
     before do
-      BookingRequests.api = api
       allow(api).to receive(:slots).and_return(
         [{ 'date' => '2017-06-02', 'start' => '0900', 'end' => '1300' }]
       )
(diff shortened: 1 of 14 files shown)


jongilbraith:
[As a booking manager who has set no availability for a location I don't want the customer to be given the 'Book online' option](https://trello.com/c/r1Zui44B/201-as-a-booking-manager-who-has-set-no-availability-for-a-location-i-dont-want-the-customer-to-be-given-the-book-online-option)

spec/lib/booking_requests/booking_requests_spec.rb
```diff
@@ -2,8 +2,16 @@
...
 
   describe '.slots' do
-    before do
+    around do |example|
+      previous_api = BookingRequests.api
       BookingRequests.api = api
+
+      example.run
+
+      BookingRequests.api = previous_api
```
benlovell: I think for this to be guaranteed it needs to be in an `ensure` since the example could fail but others still run after, so this state would leak. Also, it's duplicated below, maybe we can extract it?

jongilbraith: Good plan on both counts, will take a look.

Side note: pretty sure we're employing this pattern in a few places so we quite possibly should go search out and refactor elsewhere too.

benlovell: I've only seen it in cukes and they're extracted into support helpers that are pulled in, based on tags.

jongilbraith: I think I've used it in some feature specs and if so I've certainly not been employing `ensure`, so will just double check to see if that's the case.

PR #955, opened 2017-06-29
Permit between one and three face-to-face slots

benlovell:
Previously we required three choices at all times. We now need just one
slot chosen for the customer to proceed but still recommend up to 3
choices.

Needs coordination with guidance-guarantee-programme/planner#175.

journey


benlovell:
[As a customer when booking an appointment I am able submit the booking form even if I haven't requested three slots](https://trello.com/c/u18jebp4/160-as-a-customer-when-booking-an-appointment-i-am-able-submit-the-booking-form-even-if-i-havent-requested-three-slots)

PR #956, opened 2017-07-04
Self serve journey

jongilbraith:
The new self serve journey.

* Links from the home page ('Explore your options')
* Intro
* 6 pension type options
* 4 extra options info options (plus two compulsory included ones)
* Summary consisting of a page for each selected option, the two compulsory ones interspersed amongst the selected ones and a final 'additional info' page
* Links to print or download a pdf in the right hand sidebar
* Print/download version comes with an extra page of introductory content

Due to a more than minimal amount of logic needed to drive the way the information is presented, we weren't able to follow the same pattern as used in the calculators (custom govspeak form handlers) so this uses a dedicated controller and views, leveraging to guides for the bulk of the content.

All other content in the templates uses locale files, ready for potentially being translated to welsh.

Other changes:

* The logic for looking up guides has changed - previously it would traverse subdirectories to look for guides of the given id (a bug/oversight I would say) now it adheres to the path of the passed in id.
* The logic for generating the url for the locale switcher has been moved into a helper and expanded upon. It's rather clunky and has scope for tidying, but a cleaner overall approach isn't obvious.


jongilbraith:
[Rework content to improve customer self-serve journey](https://trello.com/c/Lbe39ViY/183-rework-content-to-improve-customer-self-serve-journey)

app/controllers/application_controller.rb
```diff
@@ -28,6 +28,22 @@ def alternate_locales
     I18n.available_locales - Array(I18n.locale)
   end
 
+  def alternate_url(new_locale, options = {}) # rubocop:disable Metrics/MethodLength
+    new_params = params.delete_if { |k| %w(utf8 authenticity_token).include?(k) }
```
benlovell: Not sure if these are `ActionController::Params` but could we use `Hash#except` here instead of the `delete_if` stuff?

app/domain/pension_summary.rb
```diff
@@ -0,0 +1,93 @@
...
+    how_my_pension_is_taxed
+  ).freeze
+
+  OPTIONS = [*PRIMARY_OPTIONS, *SECONDARY_OPTIONS, *COMPULSORY_OPTIONS].freeze
+
+  attr_accessor *STEPS
+
+  attr_accessor :current_step
+
+  def initialize(attributes)
```
benlovell: You don't use `attributes` so maybe just replace that arg with `*` and just call `super` bare?

config/routes.rb
```diff
@@ -26,6 +26,13 @@
       get 'facebook-landing', to: 'marketing#facebook'
       get 'about', to: 'marketing#about'
 
+      get 'build-your-pensions-summary', to: 'pension_summaries#start'
```
benlovell: You could cut down some of this duplication by nesting something like 

```ruby
scope '/build-your-pensions-summary', controller: 'pension_summaries' do
  get 'step-one'
  get 'step-two'
  # ...
end
```

app/controllers/application_controller.rb
jongilbraith: Yep, exactly that, [no except here](http://api.rubyonrails.org/classes/ActionController/Parameters.html) 😔

app/domain/pension_summary.rb
jongilbraith: Excellent plan 👍

config/routes.rb
jongilbraith: 👍 nice

app/controllers/application_controller.rb
benlovell: Dang. Thought as much. It's shame they don't support more of hash's interface.

PR #960, opened 2017-07-07
Pool Redis connections for locations

benlovell:
Previously Redis connections were unbounded by the location reader.
This change ensures they are monentarily checked out from a bounded
connection pool, so that during times of increased load we are not
exhausting all available Redis connections. This was particularly
evident while the location pages are crawled by various bots.


benlovell:
[Pool Redis connections](https://trello.com/c/dztAKws2/230-pool-redis-connections)

jongilbraith:
[Translate home page to welsh](https://trello.com/c/pcu1VcLF/223-translate-home-page-to-welsh)

PR #963, opened 2017-07-07
Track printing and downloading of pension summaries

jongilbraith:
By tracking clicks on the print and download buttons.


jongilbraith:
[As a Analyst I want to know how many users of the Self Serve Journey either printed or downloaded the summary](https://trello.com/c/wl9OJO9P/231-as-a-analyst-i-want-to-know-how-many-users-of-the-self-serve-journey-either-printed-or-downloaded-the-summary)

PR #967, opened 2017-07-10
Update link from pension type tool to self serve journey

jongilbraith:
A link in the pension type tool that previously linked back to the start of the tool, now linking to the self serve journey.


jongilbraith:
[Add link from p-type tool to self-serve journey when it goes live](https://trello.com/c/e1EXJuZr/217-add-link-from-p-type-tool-to-self-serve-journey-when-it-goes-live)

PR #975, opened 2017-07-15
Ensure correct locations are suggested

benlovell:
When suggesting alternate locations we should exclude those that are
not available for online bookings.


benlovell:
[Limited availability should only suggest locations that are online booking enabled](https://trello.com/c/aaVRimfk/235-limited-availability-should-only-suggest-locations-that-are-online-booking-enabled)

PR #976, opened 2017-07-19
WIP - Track clicks of alternate locations on the booking request page

jongilbraith:
In order to track effectiveness of the alternate locations prompt, a custom event to fire whenever someone clicks through from a booking request to one of the offered locations.

PR #982, opened 2017-07-25
Add a class to identify alternate location links for the purpose of GA

jongilbraith:
We can target via css with GA but there aren't really any decent structural classes that we can leverage in order to reliably target these links, so adding some dedicated classes directly to the links.

Following the BEM pattern, went with a `ga-` prefix.

### DIFF
--- a/app/views/booking_requests/step_one.html.erb
+++ b/app/views/booking_requests/step_one.html.erb
@@ -8,7 +8,10 @@
         <p>There is limited availability in this location. Other locations near <%= @booking_request.location_name %>:</p>
         <ul>
           <% @locations.each do |location| %>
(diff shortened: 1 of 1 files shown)


jongilbraith:
[As a analyst I want to know how many people are using the alternative locations links we display when a location has limited availability so I can see if they book an appointment](https://trello.com/c/SamU9NPM/226-as-a-analyst-i-want-to-know-how-many-people-are-using-the-alternative-locations-links-we-display-when-a-location-has-limited-ava)

benlovell:
@jongilbraith This supersedes #976 now right?

jongilbraith:
@benlovell yep, that other one's not needed. Will close that one off.

PR #983, opened 2017-07-25
Complaints

jongilbraith:
A formalised process and form for submission of booking-specific (or otherwise) complaints.

This is yet to be translated.

## The form

contact page

## Submitted

Redirects after submission with a param in the query string to toggle this state

contact page - submitted


jongilbraith:
[As a central support administrator I want to know as much information about the customer and their appointment when i receive a complaint so that I can easily handle it.](https://trello.com/c/mq4P8OTD/239-as-a-central-support-administrator-i-want-to-know-as-much-information-about-the-customer-and-their-appointment-when-i-receive-a)

PR #984, opened 2017-07-25
Limited availability only for online booking

benlovell:
Locations that are not enabled for online booking should never report
limited availability.


benlovell:
[Do not show 'Limited availability' message for locations which have online booking disabled](https://trello.com/c/fqoGhZdK/246-do-not-show-limited-availability-message-for-locations-which-have-online-booking-disabled)

PR #986, opened 2017-08-07
Validate feedback emails more robustly

benlovell:
We were permitting incorrectly formatted email addresses for feedback
and when these get through they're refused by ZenDesk resulting in an
error message to the customer. This change ensures we validate emails
more rigorously.


benlovell:
[Validate emails robustly when given via the feedback forms](https://trello.com/c/UCV3Z3uf/256-validate-emails-robustly-when-given-via-the-feedback-forms)

PR #989, opened 2017-08-14
Update annuity rates for August

benlovell:


benlovell:
[Update the average annuity figures used on the annuity calculator](https://trello.com/c/lPzxYGLg/265-update-the-average-annuity-figures-used-on-the-annuity-calculator)

PR #1002, opened 2017-09-14
Use canonical online booking twilio number

benlovell:
Allows administrators to provide overridden numbers for parent/child
locations.


benlovell:
[As a pension wise admin when I assign a child location an 'online booking twillio number' this is displayed in customer emails](https://trello.com/c/1xnIBIwE/273-as-a-pension-wise-admin-when-i-assign-a-child-location-an-online-booking-twillio-number-this-is-displayed-in-customer-emails)

PR #1005, opened 2017-10-17
Disable bookings for fully booked locations

benlovell:
Shows the location is fully booked when it has no availability.


benlovell:
[Disable bookings for locations with no availability](https://trello.com/c/zDVEd8iZ/314-disable-bookings-for-locations-with-no-availability)

PR #1008, opened 2017-10-24
Calculate age at date of appointment

benlovell:
This ensures that customers turning 50 by the time of their appointment
are eligible and thus can place a booking.

### DIFF
--- a/app/models/tesco/booking.rb
+++ b/app/models/tesco/booking.rb
@@ -45,7 +45,7 @@ module Tesco
     end
 
     def eligible?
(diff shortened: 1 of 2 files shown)


benlovell:
[As a Tesco employee I am able to book an appointment even though I am under 50 but will be eligible by the time I have the appointment.](https://trello.com/c/s0AUzB8g/315-as-a-tesco-employee-i-am-able-to-book-an-appointment-even-though-i-am-under-50-but-will-be-eligible-by-the-time-i-have-the-appoi)

PR #1009, opened 2017-10-31
Identify agents for customer bookings

benlovell:
Identifies TP agents and gives them a custom path when placing bookings on behalf of a customer.


benlovell:
[As a TP agent I am able to request an appointment for a customer](https://trello.com/c/WRq8UUhd/318-as-a-tp-agent-i-am-able-to-request-an-appointment-for-a-customer)

PR #1013, opened 2017-12-04
Introduce specific tags for feedback channels

benlovell:
This categories the channels when processing feedback to zendesk.


benlovell:
[As a Pension Wise Admin in Zendesk I am able to distinguish between face to face and telephone feedback](https://trello.com/c/Z9I8KFfz/350-as-a-pension-wise-admin-in-zendesk-i-am-able-to-distinguish-between-face-to-face-and-telephone-feedback)

PR #1014, opened 2017-12-05
Don't show phone number with no availability

benlovell:
When the location is enabled for online bookings and the location has
no availability the customer was always presented the call centre
number. If they called the number they would be told the location has
no availability and they would reach a dead end. This change ensures
they are made aware early on when the location has no appointment
availability.


benlovell:
[As a customer, I need to know that F2F appointments cannot be booked over the telephone when they are not visible on the website.](https://trello.com/c/S5gtWhTx/351-as-a-customer-i-need-to-know-that-f2f-appointments-cannot-be-booked-over-the-telephone-when-they-are-not-visible-on-the-website)

PR #1016, opened 2017-12-20
Include temporary new year confirmation

benlovell:
Lets the customer know not to expect a response until the new year when
placing a booking.


benlovell:
[As a customer I should be made aware of delays to the booking process over the Christmas period](https://trello.com/c/JFvvkIVS/364-as-a-customer-i-should-be-made-aware-of-delays-to-the-booking-process-over-the-christmas-period)

PR #1020, opened 2018-01-12
Update annuity rates per the latest averages

benlovell:


benlovell:
[Update averaged annuity rates](https://trello.com/c/cWCbTxNM/374-update-averaged-annuity-rates)

PR #1026, opened 2018-02-07
Spam warning

benlovell:
Mention that emails may be directed to the customer's spam folder and
that they should check there in the first instance.


benlovell:
[As a customer, I need to be prompted that the booking request and appointment confirmation email may enter my junk/spam mailbox.](https://trello.com/c/A2xsnKP7/397-as-a-customer-i-need-to-be-prompted-that-the-booking-request-and-appointment-confirmation-email-may-enter-my-junk-spam-mailbox)

PR #1028, opened 2018-02-27
Ensure postcodes are never submitted in URL's

benlovell:
This was missed previously in the last round of changes to postcode
search, in order to avoid the passing of postcodes in URL's.


benlovell:
[Amend 'appointment locations near you' and 'more details about' pages so it does not include user submitted postcodes](https://trello.com/c/WDenJXVF/415-amend-appointment-locations-near-you-and-more-details-about-pages-so-it-does-not-include-user-submitted-postcodes)

PR #1029, opened 2018-03-05
Draw attention to the potential delays

benlovell:
This is due to the recent adverse weather.

Also fixes a broken link.


benlovell:
[Add message on "Find an appointment location near you" page to inform customers that they may experience wait due to recent adverse weather conditions](https://trello.com/c/rA2HJChY/423-add-message-on-find-an-appointment-location-near-you-page-to-inform-customers-that-they-may-experience-wait-due-to-recent-advers)

PR #1030, opened 2018-03-07
Replace MAS number with the latest

benlovell:
This was out of date.


benlovell:
[Ensure summary documents contain the correct telephone number for MAS](https://trello.com/c/vKSkj155/430-ensure-summary-documents-contain-the-correct-telephone-number-for-mas)

PR #1032, opened 2018-03-12
List Tesco locations in a simple grouped index

benlovell:
Lists the locations in a first letter style index page. Clicking through
brings the customer to the details for that particular location.


benlovell:
[As a Tesco employee, I need to be able to find my store on the booking portal](https://trello.com/c/VrJjjoHH/435-as-a-tesco-employee-i-need-to-be-able-to-find-my-store-on-the-booking-portal)

PR #1034, opened 2018-03-13
Clarify messages for slot picker

benlovell:
These were causing confusion and have been reported via zendesk on
multiple occasions.


benlovell:
[Amend "You can only book an appointment 3 days in advance" message from customer booking form (Planner)](https://trello.com/c/WKNCpTXQ/436-amend-you-can-only-book-an-appointment-3-days-in-advance-message-from-customer-booking-form-planner)

PR #1035, opened 2018-03-13
Update annuity rates per the latest averages

benlovell:


benlovell:
[Update the average annuity figures used on the annuity calculator](https://trello.com/c/Cs1NRcYL/439-update-the-average-annuity-figures-used-on-the-annuity-calculator)

PR #1038, opened 2018-03-26
Ensure all Tesco hubs are disallowed

benlovell:
These should not be spidered.


benlovell:
[Ensure all Tesco hub pages are not indexed by search engines](https://trello.com/c/7L2FCnSN/454-ensure-all-tesco-hub-pages-are-not-indexed-by-search-engines)

PR #1040, opened 2018-03-27
Link to phone appointments

benlovell:
Display the link to alternatively book a phone appointment when a
location has no availability.


benlovell:
[Amend 'no availability' message for F2F locations customers so that they are directed to telephone appointments.](https://trello.com/c/r54zMb5S/459-amend-no-availability-message-for-f2f-locations-customers-so-that-they-are-directed-to-telephone-appointments)

PR #1046, opened 2018-04-11
Permit telephone journey inline feedback

benlovell:
This was validating incorrectly prior to this change.


benlovell:
[Allow 'phone booking' inline feedback](https://trello.com/c/mcOrOBu0/481-allow-phone-booking-inline-feedback)

PR #1057, opened 2018-04-18
Hide appointment link in mobile

benlovell:
This was stacking with the sidebar appointment link and since it was
titled the same it looked a bit weird.

## Before

before

## After

after


benlovell:
[Ensure that self serve journey pages are correctly reflected on mobile view](https://trello.com/c/kTOL2stZ/489-ensure-that-self-serve-journey-pages-are-correctly-reflected-on-mobile-view)

PR #1063, opened 2018-04-19
Ensure locale is included in redirect

benlovell:
When questions are submitted without answer we should redirect back
with the locale fragment in the URI.


benlovell:
[As a customer, I need to access a Welsh Language version of the Pension Type tool](https://trello.com/c/qiuWbnNC/487-as-a-customer-i-need-to-access-a-welsh-language-version-of-the-pension-type-tool)

PR #1078, opened 2018-05-28
527 age at appointment

spannerinworks:
calculate age at the time of the appointment/slot rather than the current time


benlovell:
[As a customer I am able to book an appointment even though I am under 50 but will be eligible by the time I have the appointment.](https://trello.com/c/g8XAEwvC/527-as-a-customer-i-am-able-to-book-an-appointment-even-though-i-am-under-50-but-will-be-eligible-by-the-time-i-have-the-appointment)

PR #1085, opened 2018-06-11
Guard fake server startup

benlovell:
Due to a timing issue the mailgun fake was started twice between
individual test runs this change ensures this doesn't happen.


benlovell:
[Fix flakey `shutdown` in mailgun email validation specs](https://trello.com/c/sfHxcFW2/571-fix-flakey-shutdown-in-mailgun-email-validation-specs)

PR #1086, opened 2018-06-16
Add phone hint for SMS reminders

benlovell:
Gives the customer a hint that we will send an SMS reminder when they
provide a mobile phone number.


benlovell:
[As a Tesco employee, I receive an appointment reminder SMS and can cancel via response](https://trello.com/c/P4LBDx4T/575-as-a-tesco-employee-i-receive-an-appointment-reminder-sms-and-can-cancel-via-response)

PR #1095, opened 2018-08-16
Update annuity rates

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/OSCLUEGW/644-update-average-annuity-rates)

PR #1101, opened 2018-09-05
Amend self serve pilot

pixeltrix:


pixeltrix:
[Amend self serve pilot journey](https://trello.com/c/cy9tXz5x/662-amend-self-serve-pilot-journey)

PR #1103, opened 2018-09-24
Separate further signposting for web/print

benlovell:
The printed/download version will show in full while the web version
will link to the newly introduced 'Trusted sources' page. This was
deemed necessary otherwise the final pension summary step is overloaded
with the full trusted sources content.


benlovell:
@arwelowen We've had to separate the final signposting in the pension summary, could you please translate this when you get a moment?

PR #1105, opened 2018-09-27
Link to TPAS webchat in pension type tool

benlovell:
When the customer has javascript enabled (the tool requires javascript)
we display a link to the TPAS webchat tool.


benlovell:
[As a customer of the PW service I would like to be able to chat on line to confirm my eligibility.](https://trello.com/c/0LAC0TnC/404-as-a-customer-of-the-pw-service-i-would-like-to-be-able-to-chat-on-line-to-confirm-my-eligibility)

PR #1109, opened 2018-10-02
Add env var config to disable the pilot summaries

pixeltrix:


pixeltrix:
[Pause self serve journey for one week](https://trello.com/c/AvQER8iy/706-pause-self-serve-journey-for-one-week)

PR #1113, opened 2018-10-25
Make the validation key configurable

benlovell:


benlovell:
[TP agent / TAP / Customer booking forms are showing validation error where email address is correct.](https://trello.com/c/crGhDabf/729-tp-agent-tap-customer-booking-forms-are-showing-validation-error-where-email-address-is-correct)

PR #1114, opened 2018-10-30
Disable mailgun validations at will

benlovell:
Due to the fact we sometimes hit the limit for mailgun validation API
requests and their inability to allow us to purchase more, since the
service was provisioned via heroku - we needed a way to quickly disable
customer-facing validations if necessary.

The presence of the `DISABLE_MAILGUN_VALIDATION` environment variable
will disable customer validations. Unsetting the variable will switch
them back on.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,6 +16,8 @@ module ApplicationHelper
   end
 
   def email_validation_data_attributes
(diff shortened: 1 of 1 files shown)


benlovell:
[TP agent / TAP / Customer booking forms are showing validation error where email address is correct.](https://trello.com/c/crGhDabf/729-tp-agent-tap-customer-booking-forms-are-showing-validation-error-where-email-address-is-correct)

PR #1116, opened 2018-11-13
Latest annuity rates for November

benlovell:


benlovell:
[Update Annuity Rates (November 2018)](https://trello.com/c/7mqi04i4/742-update-annuity-rates-november-2018)

PR #1119, opened 2018-12-10
Update annuity rates for December

benlovell:


benlovell:
[Update annuity rates for December](https://trello.com/c/fPaeybMI/772-update-annuity-rates-for-december)

PR #1120, opened 2018-12-12
Disable full confirmation for bookings temporarily

benlovell:
We've seen issues with confirmations and higher than usual numbers of
duplicate bookings which appear to be related. I'll disable this
confirmation while we investigate a fix.


benlovell:
[Investigate duplicate bookings](https://trello.com/c/ePZm1cQs/780-investigate-duplicate-bookings)

PR #1121, opened 2018-12-12
Ensure appointment confirmations are displayed

benlovell:
Previously we were checking an incorrect predicate to determine whether the
booking was automatically fulfilled in realtime or not. Also, this was
fragile when a customer returned to the confirmation page by refresh and
the required `flash` data was no longer present. This approach ensures
bookings and appointments are displayed correctly without error.


benlovell:
[Investigate duplicate bookings](https://trello.com/c/ePZm1cQs/780-investigate-duplicate-bookings)

PR #1127, opened 2019-01-03
Fix timing issue in CI environment

benlovell:


benlovell:
[Fix build timing issue with feedback specs](https://trello.com/c/BkqScHoU/793-fix-build-timing-issue-with-feedback-specs)

PR #1139, opened 2019-02-11
Update annuity rates for February

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/OpPl81Qt/884-update-average-annuity-rates)

PR #1140, opened 2019-02-12
Remove SFGB privacy notice

benlovell:
This is no longer needed as the transition has occurred.


benlovell:
[Remove SFGB notice on book free appointments page](https://trello.com/c/lq7TYy8Y/888-remove-sfgb-notice-on-book-free-appointments-page)

PR #1141, opened 2019-02-15
Update landing page to use apple creatives

pixeltrix:


pixeltrix:
[Update landing page with apples creative](https://trello.com/c/uCjZXbU9/886-update-landing-page-with-apples-creative)

PR #1145, opened 2019-02-27
Add environment variable to disable pilot data collection

pixeltrix:


benlovell:
[Amend self serve pilot journey (remove 'about-you' page)](https://trello.com/c/PojU3lQD/910-amend-self-serve-pilot-journey-remove-about-you-page)

PR #1147, opened 2019-03-08
Translated provider content

benlovell:


benlovell:
[As a customer, I want to know the designated providers of Pension Wise guidance so that I am aware which organisation delivers my appointment.](https://trello.com/c/CiqbSIYJ/920-as-a-customer-i-want-to-know-the-designated-providers-of-pension-wise-guidance-so-that-i-am-aware-which-organisation-delivers-my)

PR #1152, opened 2019-03-27
Update page title to alter organic search title

benlovell:
This is being implemented during a test period to assess any change on 
booking numbers when directed from organic search results.

Also updates ruby and rails dependencies.


benlovell:
[Update Pension Wise title for natural search](https://trello.com/c/7lSXeljz/948-update-pension-wise-title-for-natural-search)

PR #1154, opened 2019-03-29
Add additional campaign landing pages

pixeltrix:
These are platform and campaign specific so that there's no need to use a schedule and different campaigns can run at the same time.

### DIFF
--- a/app/controllers/marketing_controller.rb
+++ b/app/controllers/marketing_controller.rb
@@ -8,6 +8,7 @@ class MarketingController < ApplicationController
   end
   alias facebook landing
   alias about landing
+  alias campaign landing
 
   def footer?
     false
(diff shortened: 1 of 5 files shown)


pixeltrix:
[As a campaign manager, I need new landing pages for each of our biddable media sources so that I can accurately attribute telephone calls to specific media source.](https://trello.com/c/vD8nU0LY/939-as-a-campaign-manager-i-need-new-landing-pages-for-each-of-our-biddable-media-sources-so-that-i-can-accurately-attribute-telepho)

pixeltrix:
[As a campaign manager, I need new landing pages for each of our biddable media sources so that I can accurately attribute telephone calls to specific media source.](https://trello.com/c/vD8nU0LY/939-as-a-campaign-manager-i-need-new-landing-pages-for-each-of-our-biddable-media-sources-so-that-i-can-accurately-attribute-telepho)

PR #1159, opened 2019-04-10
Improve wording for submit button

benlovell:
This should not refer to booking requests now all locations are realtime
enabled and result in actual appointments.


benlovell:
[Replace 'Submit booking request' text on F2F booking form](https://trello.com/c/nxwC5UqG/957-replace-submit-booking-request-text-on-f2f-booking-form)

PR #1162, opened 2019-05-13
Update to latest annuity rates

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/gsf08RLk/987-update-average-annuity-rates)

PR #1163, opened 2019-05-30
RNIB Issue 6: Before you start placement

benlovell:
This was deemed unnecessary so can be removed.


benlovell:
[Amend Pension type tool (first page)](https://trello.com/c/5zqtKVaP/1005-amend-pension-type-tool-first-page)

PR #1165, opened 2019-06-13
Update annuity rates for June/July

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/j1iN2rsE/1019-update-average-annuity-rates)

PR #1167, opened 2019-06-26
Correct tax calculation for 2019/20

benlovell:


benlovell:
[Update personal allowance for 19/20 - take cash in chunks example](https://trello.com/c/4TfXaHsi/1032-update-personal-allowance-for-19-20-take-cash-in-chunks-example)

PR #1168, opened 2019-07-03
Don't fudge a prefix for employers

benlovell:
Previous to the employer planner changes we had to trim the 'Tesco'
identifier from a location. We now refer to this information via the
associated 'employer' so this is no longer necessary.


benlovell:
[Turn Tesco Planner into a multi-tenant application for other employers](https://trello.com/c/EBALnmNo/690-turn-tesco-planner-into-a-multi-tenant-application-for-other-employers)

PR #1169, opened 2019-07-16
Switch from MoJ calendar to our slot picker

benlovell:
Ours is a much better place in terms of usability and accessibility, plus
it brings consistency with the other journeys.


benlovell:
[Switch face-to-face calendar to phone calendar components to address 12 outstanding accessibility issues](https://trello.com/c/ASeSZpQk/1011-switch-face-to-face-calendar-to-phone-calendar-components-to-address-12-outstanding-accessibility-issues)

PR #1171, opened 2019-07-17
Update annuity rates for July/August

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/lRDloEQp/1070-update-average-annuity-rates)

PR #1181, opened 2019-08-19
404 for missing locations

benlovell:
This was erroring before but should 404 instead.


benlovell:
[Show no availability for inactive locations](https://trello.com/c/1kwAQgkZ/1084-show-no-availability-for-inactive-locations)

PR #1183, opened 2019-09-04
Improve 'unavailability' message

benlovell:
This could be misleading when a location is enabled prior to available
slots being added.


benlovell:
[Employer Planner should not show ‘fully booked’ message when no slots have been added](https://trello.com/c/JdprO5GY/1117-employer-planner-should-not-show-fully-booked-message-when-no-slots-have-been-added)

PR #1185, opened 2019-09-05
Ensure configuration values are integers

benlovell:
When these values are set they were not converted to integers and were
causing issues further along. This change ensures they are always
converted to integers regardless whether they're defaulted or come from
the configured environment variables.


benlovell:
[Increase the telephone appointments API read timeout](https://trello.com/c/aYQI6Lxh/1121-increase-the-telephone-appointments-api-read-timeout)

PR #1193, opened 2019-10-11
Don't skip CSRF for feedback

benlovell:
This was switched off across the board for the pension type tool since this
is a hybrid journey that uses static (therefore cached and needing cookies
stripped for cloudflare) content and cannot serve a CSRF token and the
requisite cookie. It's easier just to turn what was previously the inline
form into a link to the feedback form.

This change has been made to mitigate the anonymous pen testing the site 
has suffered recently.


benlovell:
[Block Nessus vulnerability scans](https://trello.com/c/BPvDXemR/1152-block-nessus-vulnerability-scans)

PR #1196, opened 2019-11-07
Update annuity rates

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/in1uyume/1183-update-average-annuity-rates)

PR #1203, opened 2019-12-17
Remove mailgun inline validation

benlovell:
This isn't robust enough to rely on for production usage.


benlovell:
[Inline email validation via Mailgun is failing](https://trello.com/c/Vqw6ltP4/1217-inline-email-validation-via-mailgun-is-failing)

PR #1204, opened 2019-12-18
Add the MAPS banner to headings

benlovell:
Adds the new MAPS header banner across the website.


benlovell:
[As a user, I want to be clearly informed of the relationship of TPAs with MAPs, so that I know where my guidance is coming from](https://trello.com/c/EdkwyN2s/1205-as-a-user-i-want-to-be-clearly-informed-of-the-relationship-of-tpas-with-maps-so-that-i-know-where-my-guidance-is-coming-from)

PR #1209, opened 2020-01-02
Update annuity rates for January

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/7nfzNxvi/1235-update-average-annuity-rates)

PR #1210, opened 2020-01-05
Bumps Ruby and gems

benlovell:
Bumps Ruby to latest in 2.6 series, Rails to latest in 5.2 series and
some other mainstream dependencies.

I had to pin a few gems to avoid breaking changes.


benlovell:
[Upgrade Ruby and gem dependencies](https://trello.com/c/zirpvT2A/559-upgrade-ruby-and-gem-dependencies)

PR #1213, opened 2020-01-23
Display access information when provided

benlovell:
When access information is assigned to a face-to-face location we ought
to display it during the customer's booking journey.


benlovell:
[As a F2F customer I want venue access information available on the Pension Wise website and automated emails so that I can access my appointment](https://trello.com/c/zE51smxr/904-as-a-f2f-customer-i-want-venue-access-information-available-on-the-pension-wise-website-and-automated-emails-so-that-i-can-acces)

PR #1215, opened 2020-02-05
Handle missing employer locations

benlovell:
Previously this scenario would display the 'technical error' page. It
should instead be handled as a 404 with the associated 'missing content'
page displayed.


benlovell:
[Don't show the "technical error" message for inactive locations](https://trello.com/c/M96j3DbV/1284-dont-show-the-technical-error-message-for-inactive-locations)

PR #1225, opened 2020-03-04
Display location with coordinates in searches

benlovell:
The actual location details page was using coordinates but the search
results page was still using the encoded address to display a location
pin on a map. This change brings them up to parity.


benlovell:
[Geocode CAS locations so they show correctly on Google Maps](https://trello.com/c/DW0Zjyld/1302-geocode-cas-locations-so-they-show-correctly-on-google-maps)

PR #1228, opened 2020-03-13
Temporarily remove inline zendesk feedback

benlovell:
Contacts via zendesk are currently failing so this will be reinstated
once the account is working correctly.


benlovell:
[Temporarily remove complaints form and in-line forms throughout website](https://trello.com/c/Z00DzlJS/1345-temporarily-remove-complaints-form-and-in-line-forms-throughout-website)

PR #1242, opened 2020-04-15
Round up state pension amounts

benlovell:
This brings parity with the GOVUK guidance on the same content.


benlovell:
[Update state pension amounts on Pension Wise website](https://trello.com/c/DMKYaL31/1395-update-state-pension-amounts-on-pension-wise-website)

PR #1246, opened 2020-04-21
Coronavirus changes for employer bookings

benlovell:
Employer appointments are facilitated via the telephone channel
temporarily due to the ongoing coronavirus crises.


benlovell:
[As a Costain employee I need to be informed that my Pension Wise appointment will be delivered by phone](https://trello.com/c/hpOrwP6R/1409-as-a-costain-employee-i-need-to-be-informed-that-my-pension-wise-appointment-will-be-delivered-by-phone)

PR #1248, opened 2020-04-27
Bump templates for signposting

benlovell:


benlovell:
[Ensure signposting list on large print summary document is formatted correctly](https://trello.com/c/i9Xm1bcN/1415-ensure-signposting-list-on-large-print-summary-document-is-formatted-correctly)

PR #1249, opened 2020-05-06
Remove homepage top links

benlovell:
An experiment concluded that self-serve completions were higher when
these links are not present in the homepage.


benlovell:
[Remove popular on box from Pension Wise homepage](https://trello.com/c/BT2REqee/1424-remove-popular-on-box-from-pension-wise-homepage)

PR #1250, opened 2020-05-14
Update signposting to tax estimate

benlovell:
Several guiders have mentioned that this would serve Scottish customers
better.


benlovell:
[Update supporting text for take whole pot and take cash in chunks calculators](https://trello.com/c/NKWdCa07/1434-update-supporting-text-for-take-whole-pot-and-take-cash-in-chunks-calculators)

PR #1256, opened 2020-06-24
Remove some of the COVID-19 signposting

benlovell:
Some of this is no longer relevant or doesn't need calling out so prominently.


benlovell:
[Amend and remove coronavirus wording on Pension Wise website](https://trello.com/c/PkVy3iRS/1462-amend-and-remove-coronavirus-wording-on-pension-wise-website)

PR #1260, opened 2020-06-29
Update annuity rates for June/July

benlovell:


benlovell:
[Update average annuity rates](https://trello.com/c/mh03ZgNt/1466-update-average-annuity-rates)

PR #1261, opened 2020-06-29
Temporarily remove inline zendesk feedback

benlovell:
Disable zendesk API interations while the account issues persist.


benlovell:
[Zendesk account is suspended](https://trello.com/c/1YLYX3LY/1467-zendesk-account-is-suspended)

PR #1267, opened 2020-07-16
Address issue with sidebar spacing

benlovell:
This was bleeding into the next row due to a recent change for a11y
since it was reading two buttons continually. This change addresses both
issues.


benlovell:
[Issue 18 NVDA: the buttons are not read or reached in isolation](https://trello.com/c/a3R2p2Cp/822-issue-18-nvda-the-buttons-are-not-read-or-reached-in-isolation)

PR #1269, opened 2020-07-21
Process extra dimensions for phone appointments

benlovell:
Phone appointments now track a 'funnel' from new to either ineligible or
complete. For the ineligible step we also pass the reason(s) for
ineligibility.


benlovell:
[Change hook on /ineligible page so that it shows the reason why appointment could not be booked](https://trello.com/c/7wEDNivA/1480-change-hook-on-ineligible-page-so-that-it-shows-the-reason-why-appointment-could-not-be-booked)

